PR #1, opened 2020-07-06
Change Travis config to support both Python and JS

rsnyman:
Update the Travis config for the latest versions of Python, change to pytest, and add JS tests


john-dupuy:
LGTM

PR #2, opened 2020-07-06
Add readthedocs config

rsnyman:


rsnyman:
The Travis check will be fixed with #1

PR #7, opened 2020-07-08
Made run-list component and env pills clickable for filtering

jjaquish:
Added clickable filtering to the component and env pills in the run list, essentially identical to that done in the result list.

### DIFF
--- a/frontend/src/run-list.js
+++ b/frontend/src/run-list.js
@@ -19,6 +19,7 @@ import { Link } from 'react-router-dom';
 
 import { Settings } from './settings';
 import {
+  buildBadge,
   buildParams,
   buildUrl,
   getActiveProject,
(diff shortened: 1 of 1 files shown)


frontend/src/run-list.js
```diff
@@ -255,6 +276,24 @@ export class RunList extends React.Component {
...
+      this.getRuns();
+      this.setState({
+        fieldSelection: null,
+        operationSelection: 'eq',
+        textFilter: '',
+        resultSelection: null,
+        runSelection: null,
+        boolSelection: null,
+        inValues: []
+      });
```
rsnyman, line 292: You only need these fields:

```javascript
fieldSelection: null,
operationSelection: 'eq',
textFilter: '',
inValues: [],
boolSelection: null,
```

PR #14, opened 2020-07-13
Fix warnings for links, limit reports to 1e6 documents

john-dupuy:
* Fixing a compilation warning for not having `noopener` in the `rel` for links. 
* Limiting reports to 1e5 documents

The latter comes from some investigation locally, running (unfiltered) `html` reports with different values for the max documents. I limited my local mongo container to 1 GiB of memory and 1 CPU. 

| REPORT_MAX_DOCUMENTS | Runtime (s) | Approx File Size of Report | Report type |
| ------------- | ------------- | ------------- | ------------- |
| 100,000  | 80 | 250 MB | HTML |
| 200,000  | 200 | 500 MB | HTML |
| 1,000,000 | 800 | 2.5 GB | HTML |
| 100,000 | 20 | 21 MB | TEXT |
| 1,000,000 | 80 | 107 MB | TEXT |

Clearly 1e6 is too high for a realistic report. 1e5 seems okay but even then the file size is pretty high. We will need to start doing cleanup of report files if people start actually using them. @rsnyman do you think we should limit it to something even lower (5e4 ?)

Clearly, the file size is dependent on the type of report that is run (i.e. text reports will be much smaller than html reports). Therefore, we should place report-type specific limits on the MAX_DOCUMENTS. However, we need more data (extend the table presented here for more report types) to decide on these limits. For now, I think it's OK to apply a global limit.

### DIFF
--- a/backend/ibutsu_server/tasks/reports.py
+++ b/backend/ibutsu_server/tasks/reports.py
@@ -31,7 +31,7 @@ FAILURE_PERC_WARN = 0.2  # % of tests failing to be warning level
 FAILURE_PERC_DANGER = 0.4
 
 REPORT_COUNT_TIMEOUT = 2.0  # timeout for counting documents in report
-REPORT_MAX_DOCUMENTS = 1000000  # max documents for reports
+REPORT_MAX_DOCUMENTS = 100000  # max documents for reports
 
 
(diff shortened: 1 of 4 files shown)


rsnyman:
Honestly, I think that 100,000 is probably fine.

PR #28, opened 2020-08-06
[Heatmap] Follow up to #27

john-dupuy:
This is a follow up PR to #27 to make it so that the heatmap only shows builds which actually exist in the DB. It uses an aggregation similar to the one used in the JJV to pull the most recent `x` builds from the DB. In so doing it removes the need for the `build_number` arg. 

Now the heatmap should always show the number of builds specified in `Set builds to:` parameter. @rsnyman can you do a bit of local testing here to make sure I didn't break anything?

Showing a similar image to the one shown in #27, you can see that the builds go all the way back to `284` now:

### DIFF
--- a/backend/ibutsu_server/widgets/jenkins_heatmap.py
+++ b/backend/ibutsu_server/widgets/jenkins_heatmap.py
@@ -1,9 +1,10 @@
 from bson import ObjectId
 from ibutsu_server.mongo import mongo
-from pymongo.errors import OperationFailure
+from pymongo import DESCENDING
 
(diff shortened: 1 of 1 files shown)


john-dupuy:
hmm I am seeing some strange behavior with some of the Jenkins runs in my local DB, going to investigate and see if a fix is necessary.

john-dupuy:
Ok I think the issue was that I was sorting by the string value `build_number` instead of the numeric `start_time` in the `_get_builds` aggregation, the issue should be sorted out now.

PR #33, opened 2020-08-12
Migrate to postgres

rsnyman:
This is currently in PoC phase, some things are migrated, others are not.

General CRUD is (or should be) working.

To Do:
- Widgets need to be migrated
- Reports need to be migrated
- <del>Tests need to be updated</del>


john-dupuy:
Awesome stuff @rsnyman!!

Some things I noticed just playing around with this locally:
* filtering needs updating, I tried filtering by component on results and runs and it didn't work (I also tried filtering by duration and hit a `TypeError`)
* Exception on the classify failures tab: `TypeError: Boolean value of this clause is not defined`
* Exception on the reports page: `TypeError: one of hex, bytes, bytes_le, fields, or int arguments must be given`

I will play around some more and update this comment as I go.

john-dupuy:
A few issues I noticed after your recent changes:
* Now I'm seeing that `jsonify` is actually causing problems, the extra quotes are making the queries return no data
    - Once I "nullified" `jsonify`, the filtered queries started to work
    - If we can get rid of it, that would be preferable as we wouldn't have to think about how to jsonify different types
* The `update_run` task completes successfully but doesn't actually update the run object, I'm not sure what's causing this yet

john-dupuy:
@rsnyman I'm going to merge this one so we can start to work on the other pieces to support the migration to postgres. 

To do:
* update reports
* update widgets
* update importer

PR #35, opened 2020-08-18
Default to 3 retries for tasks, fix some small report bugs

john-dupuy:
We've been having issues with `html` reports failing and retrying indefinitely. This PR fixes up a couple of the exceptions that we've been seeing with those tasks. 

In addition, this PR changes the default `max_retries` for the task to 3 (which is celery's [default](https://docs.celeryproject.org/en/stable/userguide/tasks.html#Task.max_retries)). 

Going forward, I think we should specify tasks that should retry more than 3 times specifically (e.g. for the `update_run` task, I set `max_retries` to 1000). I can't think of a good reason why we should retry reports and the other tasks more than 3 times. @rsnyman is there a historical reason why we've defaulted to 1000 retries for all tasks?

### DIFF
--- a/backend/ibutsu_server/tasks/importers.py
+++ b/backend/ibutsu_server/tasks/importers.py
@@ -33,7 +33,7 @@ def _create_result(tar, run_id, result, artifacts):
     return old_id
 
 
-@task()
+@task
 def run_junit_import(import_):
     """Import a test run from a JUnit file"""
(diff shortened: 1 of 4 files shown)


rsnyman:
For now, I think this is OK. Later, we might want to revert this commit and reverse the way it is done.

PR #36, opened 2020-08-25
Convert the result summary widget to psql

john-dupuy:
Need to test with a DB with more rows than I currently have, I expect we will need to use some page_size like was originally there. 

I converted this one first because I thought it'd be the easiest.

### DIFF
--- a/backend/ibutsu_server/widgets/result_summary.py
+++ b/backend/ibutsu_server/widgets/result_summary.py
@@ -1,42 +1,49 @@
-from ibutsu_server.mongo import mongo
-from pymongo import DESCENDING
+from ibutsu_server.db.base import Integer
+from ibutsu_server.db.base import session
+from ibutsu_server.db.models import Run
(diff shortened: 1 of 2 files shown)


backend/ibutsu_server/widgets/result_summary.py
```diff
@@ -1,42 +1,49 @@
...
 PAGE_SIZE = 250
 
 
 def get_result_summary(source=None, env=None, job_name=None, project=None):
     """Get a summary of results"""
-    summary = {"passed": 0, "failed": 0, "skipped": 0, "error": 0, "total": 0}
-    filters = {}
+    summary = {"error": 0, "skipped": 0, "failed": 0, "passed": 0, "total": 0}
+    query = session.query(
+        func.sum(Run.data["summary"]["errors"].cast(sqlalchemy.Integer)),
```
rsnyman, line 14: Use the `Integer` class from `ibutsu_server.db.base`

PR #39, opened 2020-08-27
Convert the jenkins heatmap to postgres

john-dupuy:
Need to check performance when we migrate prod data to postgres, particularly for getting the distinct builds. But all original functionality of the Jenkins heatmap widget is intact. 

Screenshot from 2020-08-27 13-39-06

I also added some helper functions in `filters.py` as I noticed I was repeating code a lot across the widgets.

### DIFF
--- a/backend/ibutsu_server/widgets/jenkins_heatmap.py
+++ b/backend/ibutsu_server/widgets/jenkins_heatmap.py
@@ -1,10 +1,13 @@
-from bson import ObjectId
-from ibutsu_server.mongo import mongo
-from pymongo import DESCENDING
+from ibutsu_server.db.base import Float
+from ibutsu_server.db.base import session
+from ibutsu_server.db.models import Run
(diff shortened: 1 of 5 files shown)


backend/ibutsu_server/widgets/jenkins_heatmap.py
```diff
@@ -30,106 +33,94 @@ def _calculate_slope(x_data):
...
+        .order_by(desc("build_number"))
+    )
+
+    # add filters to the query
+    query = apply_filters(query, filters, Run)
+
+    # make the query
+    query_data = query.limit(builds).all()
+
+    return [build_number[0] for build_number in query_data]
```
rsnyman, line 56: You can actually drop the `.all()` and iterate on the resultant object of `limit()` (it's actually just another `Query` object). The query is iterable, and will do smart iteration against the database.

PR #40, opened 2020-08-27
Added support for xfailed and xpassed test results

jjaquish:


rsnyman:
Just need to fix the tests :-)

PR #45, opened 2020-09-09
Postgres less json

john-dupuy:
This changes our JSON PSQL database to a relational one that has defined columns (some of which are still JSON fields). A few notes:
* Because of the foreign key constraints between related fields, deleting e.g. a run will not work, since all the run's results will have a reference to the run. Similarly, all artifacts have a reference to a result. If we want to be able to delete these items we can perhaps use DELETE CASCADE https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK
* Needs performance testing and index definitions
* Needs testing with a project defined

Requires:
* https://github.com/ibutsu/pytest-ibutsu/pull/5
* https://github.com/ibutsu/ibutsu-client-python/pull/3


rsnyman:
Just a note on the change from a float to a str/datetime object, I think we need to make sure we handle a case where the start_time is a float.

rsnyman (approved):

PR #47, opened 2020-09-28
Update the migration script to work with new psql db

john-dupuy:
Leaving as a draft until we figure out the indexes we need.

Some things I considered: 
* Only migrating run, result, and artifact files from the most recent 3 months (determined by the newest record in the MongoDB)
  - This will keep our initial DB size small and also vastly reduce the time to run the migration script
  - If we were to migrate everything it would take ~20 hrs I believe
* Adding a `upload_date` field to the artifact model
* Only migrating artifacts, we don't have many/any recent reports, so I thought it best to just migrate artifacts
* Not migrating any imports

I got many of the column indexes, but there are still the JSON indexes to consider. 

@rsnyman I'd figure I make this part of ibutsu-server so that we can use the models' `from_dict` method that we have defined in the db models. 

The pre-commit update is because I hit https://github.com/PyCQA/pycodestyle/issues/373

### DIFF
--- a/backend/ibutsu_server/db/models.py
+++ b/backend/ibutsu_server/db/models.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from uuid import uuid4
 
 from ibutsu_server.db.base import Boolean
@@ -65,14 +66,15 @@ class FileMixin(ModelMixin):
 
 class Artifact(Model, FileMixin):
     __tablename__ = "artifacts"
-    result_id = Column(PortableUUID(), ForeignKey("results.id"), nullable=False)
-    filename = Column(Text)
+    result_id = Column(PortableUUID(), ForeignKey("results.id"), nullable=False, index=True)
+    filename = Column(Text, index=True)
     data = Column(mutable_json_type(dbtype=PortableJSON(), nested=True))
+    upload_date = Column(DateTime, default=datetime.utcnow, index=True)
 
 
 class Group(Model, ModelMixin):
     __tablename__ = "groups"
-    name = Column(Text)
+    name = Column(Text, index=True)
     projects = relationship("Project")
     data = Column(mutable_json_type(dbtype=PortableJSON(), nested=True))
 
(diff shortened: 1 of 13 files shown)


john-dupuy:
The foreign key constraint is causing some problems with bad data in my local database, I don't think we'll hit the same issues on prod (especially if we're ignoring the old stuff) but I just pulled the latest DB dump to test out tomorrow.

backend/ibutsu_server/db/models.py
```diff
@@ -68,6 +68,7 @@ class Artifact(Model, FileMixin):
     result_id = Column(PortableUUID(), ForeignKey("results.id"), nullable=False)
     filename = Column(Text)
     data = Column(mutable_json_type(dbtype=PortableJSON(), nested=True))
+    upload_date = Column(DateTime)
```
rsnyman, line 71: If you need to make a default value for this, you can do the following:

```suggestion
    upload_date = Column(DateTime, default=datetime.utcnow)
```

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -0,0 +1,296 @@
...
+    "run_id",
+    "runId",
+    "project",
+    "run",
+]
+
+# indexes for the tables
+INDEXES = {
+    "results": [
+        "CREATE INDEX results_env ON results(env);",
```
rsnyman, line 70: Indexes on individual columns can (and should) be created on the models. For example:

```python
class Result(Model):
    env = Column(String, index=True)
```

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -0,0 +1,296 @@
...
+def migrate_table(collection, Model, vprint):
+    """Migrate a collection from MongoDB into a table in PostgreSQL"""
+    # TODO: update indexes once we know them
+    conn = Base.metadata.bind.connect()
+    for sql_index in INDEXES.get(Model.__tablename__, []):
+        vprint(".", end="")
+        conn.execute(sql_index)
+
+    # for runs and results, sort by descending start_time
+    if Model.__tablename__ == "runs" or Model.__tablename__ == "results":
```
rsnyman, line 147:
```suggestion
    if Model.__tablename__ in ["runs", "results"]:
```

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -0,0 +1,296 @@
...
+                else:
+                    row["metadata"][field] = convert_objectid_to_uuid(row["metadata"][field])
+
+        obj = Model.from_dict(**row)
+        session.add(obj)
+        if idx % ROWS_TO_COMMIT_AT_ONCE == 0:
+            session.commit()
+    session.commit()
+    # at the end of the session do a little cleanup
+    if Model.__tablename__ == "runs" or Model.__tablename__ == "results":
```
rsnyman, line 199:
```suggestion
    if Model.__tablename__ in ["runs", "results"]:
```

rsnyman (changes requested):
I made a couple suggestions :-)

rsnyman (approved):

john-dupuy:
OK @rsnyman I think this is finally in the "final" state. I applied a similar logic for the artifact files that is done for the results so that we can avoid any foreign key constraint errors (i.e. getting a list of migrated results and searching for artifacts within that result list).

rsnyman (approved):

PR #54, opened 2020-10-08
Update the template

rsnyman:


john-dupuy (approved):
LGTM, only thing I'm not sure about is why the PVC's didn't get created from the template. They look right to me.

PR #57, opened 2020-10-09
Limit runs from which we determine the jenkins jobs

john-dupuy:
This improves performance of the Jenkins Job View and the heatmap. 

Also add empty and error state to JJV


rsnyman (approved):
Looks alright to me.

PR #58, opened 2020-10-09
Migration limits

rsnyman:
- Use `iso8601.parse_date` (`fromisoformat()` is only 3.7+)
- Make migration limits configurable

### DIFF
--- a/backend/ibutsu_server/scripts/mongo2postgres.py
+++ b/backend/ibutsu_server/scripts/mongo2postgres.py
@@ -19,6 +19,7 @@ from ibutsu_server.db.models import Project
 from ibutsu_server.db.models import Result
 from ibutsu_server.db.models import Run
 from ibutsu_server.db.models import WidgetConfig
+from iso8601 import parse_date
 from pymongo import MongoClient
 from sqlalchemy import create_engine
 from sqlalchemy.ext.declarative import declarative_base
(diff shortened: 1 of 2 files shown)


backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -316,17 +327,21 @@ def migrate_tables(mongo, vprint, migrate_files=False):
...
+            )
         elif collection == "results":
             # migrate in chunks of 100 runs at a time
             run_chunks = [run_ids[i : i + 100] for i in range(0, len(run_ids), 100)]
             for run_list in run_chunks:
                 result_filter = {"metadata.run": {"$in": run_list}}  # filter on runs we know exist
                 result_ids.extend(
-                    migrate_table(mongo[collection], model, vprint, filter_=result_filter)
+                    migrate_table(
+                        mongo[collection], model, vprint, limit=record_limit, filter_=result_filter
```
john-dupuy, line 340: Because the results are migrated in chunks of 100 runs at a time, the record_limit will probably never be applied here.

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -337,10 +352,20 @@ def migrate_tables(mongo, vprint, migrate_files=False):
                 for result_list in result_chunks:
                     artifact_filter.update({"metadata.resultId": {"$in": result_list}})
                     migrate_file(
-                        GridFSBucket(mongo, collection), model, vprint, filter_=artifact_filter
+                        GridFSBucket(mongo, collection),
+                        model,
+                        vprint,
+                        limit=record_limit,
```
john-dupuy, line 358: Same here about the `record_limit`

john-dupuy (approved):
Just 1 note about the `record_limit`, once the script was updated to migrate results/artifacts in chunks, it isn't really valid.

john-dupuy (approved):

PR #59, opened 2020-10-09
Estimate counts in large tables by using rows returned from Explain

john-dupuy:
Various utility functions to aid in counting large tables.
 
This makes use of (a variant on) the count estimate function described here: 
https://wiki.postgresql.org/wiki/Count_estimate

Via implementing an `Explain` that can be used to `Explain` general SQLAlchemy queries, described here: http://www.wmmi.net/documents/SQLAlchemy.pdf


rsnyman (approved):
Nice!

PR #60, opened 2020-10-12
Further performance improvements

john-dupuy:
Fetching 1000 runs is taking too long and affecting other pages in the app.

Also including some minor improvements for the heatmap and the run/results tables.

### DIFF
--- a/frontend/src/result-list.js
+++ b/frontend/src/result-list.js
@@ -374,7 +374,7 @@ export class ResultList extends React.Component {
   }
 
   getRuns() {
-    fetch(buildUrl(Settings.serverUrl + '/run', {pageSize: 20, estimate: true}))
+    fetch(buildUrl(Settings.serverUrl + '/run', {pageSize: 1000, estimate: true}))
       .then(response => response.json())
       .then(data => {
         const runs = data.runs.map((run) => run.id);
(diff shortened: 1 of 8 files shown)


frontend/src/result-list.js
```diff
@@ -374,7 +374,7 @@ export class ResultList extends React.Component {
   }
 
   getRuns() {
-    fetch(buildUrl(Settings.serverUrl + '/run', {pageSize: 1000}))
+    fetch(buildUrl(Settings.serverUrl + '/run', {pageSize: 20, estimate: true}))
```
rsnyman, line 377: So, this fetches the list of runs so that the dropdown on the results list page can show and filter runs. I can't remember if we do a server-side lookup for the list, but limiting this to 20 might be a problem.

john-dupuy: Oh right, I think we should modify this dropdown so it can accept run_id's that aren't "pre-fetched". 

Furthermore, this dropdown is from before the runs table right? I don't think we should make an API call for 1000 runs when it is easier and more user friendly to use the runs table.

john-dupuy: I reverted this change (but kept the `estimate`), we can improve this API call, potentially with on-the-fly filtering in a future PR.

PR #62, opened 2020-10-14
Make it all run at the same time

rsnyman:


backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -69,27 +72,23 @@
...
+    "ON results((data->>'classification'));",
+    "CREATE INDEX IF NOT EXISTS ix_results_assignee " "ON results((data->>'assignee'));",
+    "CREATE INDEX IF NOT EXISTS ix_results_exception_name "
+    "ON results((data->>'exception_name'));",
+    "CREATE INDEX IF NOT EXISTS ix_runs_jenkins_job_name "
+    "ON runs((data->'jenkins'->>'job_name'));",
+    "CREATE INDEX IF NOT EXISTS ix_runs_jenkins_build_number "
+    "ON runs((data->'jenkins'->>'build_number'));",
+    "CREATE INDEX IF NOT EXISTS ix_runs_jjn_jbn "
+    "ON runs((data->'jenkins'->>'build_number'), (data->'jenkins'->>'job_name'))",
```
john-dupuy, line 90: We should perhaps add one additional index for the runs:
```
CREATE INDEX IF NOT EXISTS ix_runs_summary ON runs USING gin (summary)
```

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -9,13 +9,16 @@
...
+from multiprocessing import Pool
 from uuid import UUID
 
 from bson import ObjectId
 from gridfs import GridFSBucket
 from ibutsu_server.db.models import Artifact
 from ibutsu_server.db.models import Group
 from ibutsu_server.db.models import Project
+from ibutsu_server.db.models import Report
+from ibutsu_server.db.models import ReportFile
```
john-dupuy, line 21: I think `iso8601` may need to be added to the setup.cfg file.

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -283,8 +298,72 @@ def migrate_file(collection, Model, vprint, limit=None, filter_=None):
...
+    vprint("Migrating widgets ", end="")
+    migrate_table(mongo.widgetConfigs, WidgetConfig, vprint)
+
+
+def migrate_projects(mongo, vprint):
+    """Migrate the projects"""
+    vprint("Migrating projects ", end="")
+    for mproject in mongo.projects.find():
+        vprint(".", end="")
+        pproject = Project.query.get(convert_objectid_to_uuid(mproject["_id"]))
```
john-dupuy, line 340: When running locally I got an exception about `context` on this line. I was able to resolve it by using `session.query`.

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -283,8 +298,72 @@ def migrate_file(collection, Model, vprint, limit=None, filter_=None):
...
+    migrate_table(mongo.reports, Report, vprint)
+    vprint("Migrating files for reports ", end="")
+    migrate_file(GridFSBucket(mongo, "reportFiles"), ReportFile, vprint, filter_=filter_)
+
+
+def migrate_widgets(mongo_url, mongo_db, postgres_url, is_verbose):
+    """Migrate all the widgets for the dashboards"""
+    mongo, _, vprint = setup_objects(mongo_url, mongo_db, postgres_url, is_verbose)
+    vprint("Migrating widgets ", end="")
+    migrate_table(mongo.widgetConfigs, WidgetConfig, vprint)
```
john-dupuy, line 332:
```suggestion
    migrate_table(mongo.widgetConfig, WidgetConfig, vprint)
```

backend/ibutsu_server/scripts/mongo2postgres.py
rsnyman: Sure! I'll add that.

backend/ibutsu_server/scripts/mongo2postgres.py
rsnyman: Well, it's only needed for this script, so I am not sure if we should put it in there.

backend/ibutsu_server/scripts/mongo2postgres.py
rsnyman: Yeah, cause the models in the server need the application (Flask) context. The code I originally wrote did not use the models from the server.

backend/ibutsu_server/scripts/mongo2postgres.py
rsnyman: Ah, thanks!

backend/ibutsu_server/scripts/mongo2postgres.py
john-dupuy: Makes sense, my concern was if we wanted to run this from OCP we would need it there.

backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -283,89 +280,121 @@ def migrate_file(collection, Model, vprint, limit=None, filter_=None):
...
+
+        # Now create a multiprocessing pool (it automatically uses all processors available)
+        pool = Pool(processors or None)
+        # Add the widet and reports migration functions to the pool
+        pool.apply_async(migrate_widgets, (mongo_url, mongo_db, postgres_url, is_verbose))
+        pool.apply_async(
+            migrate_reports, (mongo_url, mongo_db, postgres_url, is_verbose, table_filter)
+        )
+        # Get a list of all the runs, and add a migration job for each run to the pool
+        for run in mongo.runs.find(table_filter):
```
john-dupuy, line 383: In case something goes wrong during the migration, I think we should reverse sort these so that the more recent results are migrated first.

john-dupuy (approved):
LGTM

PR #84, opened 2020-10-29
Sort recent builds by start_time rather than build_number

john-dupuy:
This should get rid of older builds showing up in the heatmap shortly after the Jenkins migration. 

Also filtering by `min_start_time` so that we don't get builds that are "old".

Note: we should potentially add `build_url` as a parameter in the heatmap to avoid this issue altogether. But this will be a larger effort as we'll need an index on that field.


jjaquish:
LGTM

PR #95, opened 2020-11-18
Prepare for configurable dashboards

rsnyman:
- Add model for dashboards
- Add dashboard_id to widget configs
- Create an upgrade system
- Add an initial upgrade to add dashboard_id column
- Add users for future use

### DIFF
--- a/backend/ibutsu_server/db/models.py
+++ b/backend/ibutsu_server/db/models.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from uuid import uuid4
 
+from ibutsu_server.auth import bcrypt
 from ibutsu_server.db.base import Boolean
 from ibutsu_server.db.base import Column
 from ibutsu_server.db.base import DateTime
@@ -11,16 +12,27 @@ from ibutsu_server.db.base import Integer
 from ibutsu_server.db.base import LargeBinary
 from ibutsu_server.db.base import Model
 from ibutsu_server.db.base import relationship
+from ibutsu_server.db.base import Table
 from ibutsu_server.db.base import Text
 from ibutsu_server.db.types import PortableJSON
 from ibutsu_server.db.types import PortableUUID
 from ibutsu_server.util import merge_dicts
+from sqlalchemy.exc import DBAPIError
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy_json import mutable_json_type
 
 
 def _gen_uuid():
     """Generate a UUID"""
     return str(uuid4())
+
+
+users_projects = Table(
+    "users_projects",
+    Column("user_id", PortableUUID(), ForeignKey("users.id"), primary_key=True),
+    Column("project_id", PortableUUID(), ForeignKey("projects.id"), primary_key=True),
+)
 
 
 class ModelMixin(object):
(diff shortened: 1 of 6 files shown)


backend/ibutsu_server/db/models.py
```diff
@@ -164,7 +177,66 @@ class WidgetConfig(Model, ModelMixin):
     navigable = Column(Boolean, index=True)
     params = Column(mutable_json_type(dbtype=PortableJSON()))
     project_id = Column(PortableUUID(), ForeignKey("projects.id"), index=True)
```
john-dupuy, line 179: This is just the preparation right? This relation will be removed in a later PR?

backend/ibutsu_server/db/models.py
```diff
@@ -164,7 +177,66 @@ class WidgetConfig(Model, ModelMixin):
...
     weight = Column(Integer, index=True)
     widget = Column(Text, index=True)
+
+
+class User(Model, ModelMixin):
+    __tablename__ = "users"
+    email = Column(Text, index=True, nullable=False, unique=True)
+    password = Column(Text, nullable=False)
+    name = Column(Text)
+    project_id = Column(PortableUUID(), ForeignKey("projects.id"), index=True)
```
john-dupuy, line 192: I expect a user could be part of many projects. I'm thinking many-to-many here? Or getting rid of the project_id and just having `group_id`?

backend/ibutsu_server/db/models.py
```diff
@@ -164,7 +177,66 @@ class WidgetConfig(Model, ModelMixin):
...
+    group_id = Column(PortableUUID(), ForeignKey("groups.id"), index=True)
+    dashboards = relationship("Dashboard")
+
+
+class Meta(Model):
```
john-dupuy, line 197: Could you add a small docblock about this class?

backend/ibutsu_server/db/models.py
```diff
@@ -164,7 +177,66 @@ class WidgetConfig(Model, ModelMixin):
...
+    __tablename__ = "meta"
+    key = Column(Text, primary_key=True, nullable=False, index=True)
+    value = Column(Text)
+
+
+def upgrade_db(session, upgrades):
+    """Upgrade the database using Alembic
+
+    :param session: An SQLAlchemy Session object
+    :param metadata: An SQLAlchemy Metadata object
```
john-dupuy, line 207: Can this be removed?

backend/ibutsu_server/db/models.py
rsnyman: Yeah, I just left that in there for now. Later on we'll need to do some data migration and move those widgets to a "Default" or some other dashboard.

backend/ibutsu_server/db/models.py
rsnyman: Good point, I'll set that up.

backend/ibutsu_server/db/models.py
```diff
@@ -164,7 +177,66 @@ class WidgetConfig(Model, ModelMixin):
...
     title = Column(Text, index=True)
     type = Column(Text, index=True)
     weight = Column(Integer, index=True)
     widget = Column(Text, index=True)
+
+
+class User(Model, ModelMixin):
+    __tablename__ = "users"
+    email = Column(Text, index=True, nullable=False, unique=True)
+    password = Column(Text, nullable=False)
```
john-dupuy, line 190: From a cursory google search, thought on using `bcrypt` for storing passwords? 
https://uniwebsidad.com/libros/explore-flask/chapter-12/storing-passwords

backend/ibutsu_server/db/models.py
rsnyman: Converted this to a M-N relationship.

backend/ibutsu_server/db/models.py
rsnyman: Yup, we  can. I think I'll leave the actual implementation until we implement auth though.

john-dupuy: Yup, that sounds good to me.

rsnyman: I did it anyway :-)

john-dupuy (approved):
LGTM :tada:

PR #96, opened 2020-11-19
Build a new dashboard API

rsnyman:
- Add the dashboard API to the OpenAPI spec
- Build the dashboard controller
- Update the OpenAPI spec to match the API (examples, field names, etc)

Needs #95 merged first.

### DIFF
--- a/backend/ibutsu_server/controllers/dashboard_controller.py
+++ b/backend/ibutsu_server/controllers/dashboard_controller.py
@@ -1,0 +1,90 @@
+import connexion
+from ibutsu_server.db.base import session
+from ibutsu_server.db.models import Dashboard
+
+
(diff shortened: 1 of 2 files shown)


rsnyman:
Just a note, when I rebased, I overwrote your changes to `openapi.yaml` because I had co-ordinated all the IDs, and I didn't feel like tracking all of those down again

john-dupuy (commented):
LGTM, thoughts on having a `Delete` endpoint for dashboards? We have one for widget configs, so it might make sense.

rsnyman:
> LGTM, thoughts on having a `Delete` endpoint for dashboards? We have one for widget configs, so it might make sense.

I'll add that to the dashboard UI branch.

john-dupuy (approved):

PR #99, opened 2020-11-20
Dashboard frontend

rsnyman:
- Add a dropdown to select dashboard
- Update dashboard with widgets based on selected dashboard
- Add delete to the dashboard API

Please note:

- When the project is changed, there's some logic to check if the selected dashboard is part of the project
- The delete API method will unset the dashboard id on widget configs, but should we rather delete the widget configs?

### DIFF
--- a/backend/ibutsu_server/controllers/dashboard_controller.py
+++ b/backend/ibutsu_server/controllers/dashboard_controller.py
@@ -1,6 +1,7 @@
 import connexion
 from ibutsu_server.db.base import session
 from ibutsu_server.db.models import Dashboard
+from ibutsu_server.db.models import WidgetConfig
 
 
 def add_dashboard(dashboard=None):
@@ -88,3 +89,23 @@ def update_dashboard(id_, dashboard=None):
     session.add(dashboard)
     session.commit()
     return dashboard.to_dict()
+
+
+def delete_dashboard(id_):
+    """Deletes a dashboard
+
+    :param id: ID of the dashboard to delete
+    :type id: str
+
+    :rtype: tuple
+    """
+    dashboard = Dashboard.query.get(id_)
+    if not dashboard:
+        return "Not Found", 404
+    else:
+        widget_configs = WidgetConfig.query.filter(WidgetConfig.dashboard_id == dashboard.id).all()
+        for widget_config in widget_configs:
+            session.delete(widget_config)
+        session.delete(dashboard)
+        session.commit()
+        return "OK", 200
--- a/backend/ibutsu_server/openapi/openapi.yaml
+++ b/backend/ibutsu_server/openapi/openapi.yaml
@@ -759,7 +759,15 @@ paths:
       - description: Filter dashboards by project ID
         explode: true
         in: query
-        name: projectId
+        name: project_id
+        required: false
+        schema:
+          type: string
+        style: form
+      - description: Filter dashboards by user ID
+        explode: true
+        in: query
+        name: user_id
         required: false
         schema:
           type: string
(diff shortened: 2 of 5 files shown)


backend/ibutsu_server/controllers/dashboard_controller.py
```diff
@@ -51,9 +54,10 @@ def get_dashboard_list(project_id=None, user_id=None, page=1, page_size=25):
     """
     query = Dashboard.query
     if project_id:
+        print(project_id)
```
john-dupuy, line 57: Can we remove this print statement?

backend/ibutsu_server/controllers/dashboard_controller.py
```diff
@@ -51,9 +54,10 @@ def get_dashboard_list(project_id=None, user_id=None, page=1, page_size=25):
...
         query = query.filter(Dashboard.project_id == project_id)
     if user_id:
-        query = query.filter(Dashboard.user_id == user_id)
+        query = query.filter(Dashboard.user_id == UUID(user_id))
```
john-dupuy, line 60: Just curious, but why is this necessary here but not on L58? 
```
        query = query.filter(Dashboard.project_id == project_id)
```

john-dupuy:
@rsnyman Playing around with this locally, I realized that the existing widgets cannot take custom filters yet, before dashboards are released we should do that. 

Also, it is cumbersome to have to add each widget config one by one for a dashboard. I think we should have an endpoint (or potentially just add it to the existing POST endpoint) that allows you to pass a list of widgets that the dashboard will contain and a common `filter_` for the widgets, once the widgets can take a custom filter.

backend/ibutsu_server/controllers/dashboard_controller.py
rsnyman: Actually, I'm not sure. I vaguely remember adding that line when trying to debug some issues, but I think it's probably not actually necessary. I'll remove it and see if I have any problems.

rsnyman:
@john-dupuy 

1. Yeah, I thought about custom filters, and I realised that most widgets are pretty specific with their parameters, and I'm not sure how well custom filters would work with them. So for now I just left it as a link between a set of widgets and a dashboard.

2. My end goal is actually to allow users to create their own dashboards in the frontend, and I think that will remove the tediousness from interacting with the API. This is just a start, an MVP if you will, to give people a feature that they want.

backend/ibutsu_server/controllers/dashboard_controller.py
rsnyman: Whoops, thanks for spotting that.

john-dupuy:
Playing around with this, here are a couple issues I noticed: 
1) With no dashboard selected, while on the dashboard page, selecting a project from the dropdown has no effect until the page is manually refreshed
2) The active dashboard is not remembered if a user navigates to a separate page in Ibutsu.

backend/ibutsu_server/controllers/dashboard_controller.py
```diff
@@ -88,3 +89,24 @@ def update_dashboard(id_, dashboard=None):
...
+
+    :rtype: tuple
+    """
+    dashboard = Dashboard.query.get(id_)
+    if not dashboard:
+        return "Not Found", 404
+    else:
+        widget_configs = WidgetConfig.query.filter(WidgetConfig.dashboard_id == dashboard.id).all()
+        for widget_config in widget_configs:
+            widget_config.dashboard_id = None
```
john-dupuy, line 108: Since by design a widget config should be aligned with a single dashboard, I do think we should delete the widget configs when the dashboard is deleted.

rsnyman: I was worried about people potentially deleting widgets that they don't want to delete, but that's more of a UI issue that should be handled in the frontend.

rsnyman:
> 1. With no dashboard selected, while on the dashboard page, selecting a project from the dropdown has no effect until the page is manually refreshed

OK, I think I fixed this.

> 2. The active dashboard is not remembered if a user navigates to a separate page in Ibutsu.

Whoo boy. This picked up a few corner cases. Busy working on these.

john-dupuy (approved):
Nice work! This LGTM :+1:

PR #100, opened 2020-12-02
Create a UI to add dashboards and widgets

rsnyman:


backend/ibutsu_server/constants.py
```diff
@@ -34,7 +34,12 @@
                 "description": "The field in a result to group by",
                 "type": "string",
             },
-            {"name": "sort_field", "description": "The field to sort results by", "type": "string"},
+            # Apparently the widget doesn't support this field
```
john-dupuy, line 37: Oh yeah, we can remove this comment now, I got rid of that field with during the postgres migration I believe.

john-dupuy (changes requested):
Code LGTM -- this is really awesome @rsnyman! Nice work! After some local testing here are some things I noticed. Most are not bugs.

1. When no project is set and I tried to create a widget I got: ```"http://localhost:8080/api/widget-config validation error: None is not of type 'string' - 'project_id'```
2. For a future PR, it might be useful to have a button to delete a dashboard
3. When you a new dashboard, it should automatically change to the new dashboard
4. When no active dashboard is set, widgets from every dashboard appear. We should only show a few default widgets, otherwise we'll be slamming our DB with way too many widget queries. 
5. The view widgets, e.g. `jenkins-job-view` and `jenkins-analysis-view` shouldn't show up in the list of widgets to add to the dash.
6. jenkins-bar-chart and jenkins-line-chart need descriptions and some fixes

Of these, I think only 1,3 (maybe), 4, and 5 should be fixed in this PR. The rest can be done as a followup.

rsnyman:
> 1. When no project is set and I tried to create a widget I got: "http://localhost:8080/api/widget-config validation error: None is not of type 'string' - 'project_id'

I'm wondering if we shouldn't start forcing a project, and also a dashboard. We'll need to update our widgets on prod to set a dashboard ID and a project ID, but that's a simple fix.

> 2. For a future PR, it might be useful to have a button to delete a dashboard

Yup, that was my intention. Perhaps even editing widgets. But I wanted to get something out first :-)

> 3. When you a new dashboard, it should automatically change to the new dashboard

Easy thing to do :-)

> 4. When no active dashboard is set, widgets from every dashboard appear. We should only show a few default widgets, otherwise we'll be slamming our DB with way too many widget queries.

See my comment on 1.

> 5. The view widgets, e.g. jenkins-job-view and jenkins-analysis-view shouldn't show up in the list of widgets to add to the dash.

I need to check prod and see if there's a way we can do this. I couldn't immediately see on my local dev instance.

> 6. jenkins-bar-chart and jenkins-line-chart need descriptions and some fixes

The descriptions are set in the API, so we need to update them there. What fixes do we need?

john-dupuy:
@rsnyman forcing a project and dashboard makes sense to me, I say we do that. 

For 5) perhaps we can add a `type_` parameter to the widget that is defined in `constants.py`?

6) I can do in a follow up, just need to add a description to those in `constants.py` and check to see what's going on there. The widgets simply didn't display on the dashboard when I tried it.

rsnyman:
@john-dupuy OK, fixed things!

> 1. When no project is set and I tried to create a widget I got: "http://localhost:8080/api/widget-config validation error: None is not of type 'string' - 'project_id'
> 4. When no active dashboard is set, widgets from every dashboard appear. We should only show a few default widgets, otherwise we'll be slamming our DB with way too many widget queries.

The UI now forces you to select a project and then select a dashboard in order to view widgets. We'll need to update the current widgets when we deploy this.

> 3. When you a new dashboard, it should automatically change to the new dashboard

Done.

> 5. The view widgets, e.g. jenkins-job-view and jenkins-analysis-view shouldn't show up in the list of widgets to add to the dash.

OK, I added a "type" in the API and a filter as well, now the widget types are filterable.

> 6. jenkins-bar-chart and jenkins-line-chart need descriptions and some fixes

I added basic (probably wrong) descriptions to those two widgets.

frontend/src/components/new-widget-wizard.js
```diff
@@ -0,0 +1,321 @@
...
+    let selectedTypeId = event.currentTarget.value;
+    let selectedType = null;
+    let params = {};
+    this.state.widgetTypes.forEach(widgetType => {
+      if (widgetType.id === selectedTypeId) {
+        selectedType = widgetType;
+        widgetType.params.forEach(param => {
+          let paramDefault = '';
+          if (param.type === 'integer' || param.type === 'float') {
+            paramDefault = 0;
```
john-dupuy, line 119: For `weight` it makes sense to have this be `0` but for the other parameters, `0` will return no data. I did this by accident just testing it and I think other people might too.

john-dupuy (approved):
Just one small comment about the default value for `integer`/`float` parameters

backend/ibutsu_server/controllers/widget_controller.py
```diff
@@ -23,14 +23,18 @@
...
 
-def get_widget_types():
+def get_widget_types(type_=None):
     """Get the types of widgets that are available
 
     :rtype: list
     """
-    page_size = len(WIDGET_TYPES.keys())
+    widget_types = WIDGET_TYPES.values()
+    print(widget_types)
```
john-dupuy, line 32: Can we remove this `print`?

rsnyman: Doh!

frontend/src/components/new-widget-wizard.js
rsnyman: I think that in future sometime we should add default values to parameters.

john-dupuy: that sounds good

PR #105, opened 2020-12-04
Add custom filters for some widgets

john-dupuy:
This PR allows users to specify custom filters for a few widgets, so that they can filter on specific metadata. 

e.g. filtering by `platform-experience` tag
Screenshot from 2020-12-04 13-42-09

### DIFF
--- a/backend/ibutsu_server/constants.py
+++ b/backend/ibutsu_server/constants.py
@@ -39,6 +39,12 @@ WIDGET_TYPES = {
                 "description": "Count skips against the pass rate",
                 "type": "boolean",
             },
+            {
+                "name": "additional_filters",
+                "description": "Comma-separated list of additional filters, e.g. "
+                "'metadata.tags=platform-experience,metadata.assignee=username'",
+                "type": "string",
+            },
         ],
         "type": "widget",
     },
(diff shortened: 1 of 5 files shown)


backend/ibutsu_server/constants.py
```diff
@@ -39,6 +39,12 @@
                 "description": "Count skips against the pass rate",
                 "type": "boolean",
             },
+            {
+                "name": "custom_filter",
```
rsnyman, line 43: Not just `filter` ?

backend/ibutsu_server/widgets/jenkins_heatmap.py
```diff
@@ -35,8 +35,10 @@ def _calculate_slope(x_data):
     return slope
 
 
-def _get_builds(job_name, builds, project=None):
+def _get_builds(job_name, builds, project=None, custom_filter=None):
```
rsnyman, line 38: Oh, I see why you used `custom_filter`. Technically we could translate it (`filter` => `filter_`) just before we `**kwargs` it. What do you think?

rsnyman (changes requested):
Overall I like it, just some queries about `filter` vs `custom_filter`

backend/ibutsu_server/widgets/jenkins_heatmap.py
john-dupuy: Yeah I decided to call it `custom_filter` to differentiate it from the other parameters, which are really just specific filters. Maybe `additional_filters` would be more appropriate?

rsnyman: Sounds good to me.

rsnyman (approved):

PR #115, opened 2020-12-11
Remember tab + updates

rsnyman:
- On the Run and Result pages, "remember" the selected tab via the URL
- Update PatternFly React
- Make tabs "boxed"
- Fix up some style issues

### DIFF
--- a/frontend/src/components/result.js
+++ b/frontend/src/components/result.js
@@ -59,6 +59,8 @@ export class ResultView extends React.Component {
     resultId: PropTypes.string,
     hideSummary: PropTypes.bool,
     hideTestObject: PropTypes.bool,
+    history: PropTypes.object,
+    location: PropTypes.object
   }
 
   constructor(props) {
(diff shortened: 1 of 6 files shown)


frontend/src/components/result.js
```diff
@@ -175,9 +215,9 @@ export class ResultView extends React.Component {
     return (
       <React.Fragment>
         {this.state.testResult &&
-        <Tabs activeKey={this.state.artifactsTabKey} onSelect={this.handleArtifactsTabClick}>
+        <Tabs activeKey={activeTab} onSelect={this.onTabSelect} isBox>
```
john-dupuy, line 218: Can we also add `isBox` to the `Tabs` on the Jenkins Job analysis view?
https://github.com/ibutsu/ibutsu-server/blob/master/frontend/src/views/jenkinsjobanalysis.js#L210

john-dupuy (approved):
Nice enhancement! Just one small comment.

frontend/src/components/result.js
rsnyman: Do you want me to implement the tab remembering on this page too?

john-dupuy: If you'd like to. We can always do it as a follow-up.

rsnyman: Might as well :-)

rsnyman: Done :-)

PR #116, opened 2020-12-11
Pass project along with file when importing via the UI (closes #110)

### LINKED ISSUE #110: UI jUnit Import is not setting project and showing undefined on the notification
Steps to reproduce:

1) Select a project to make it active
2) Import a jUnit XML file (here you can see the undefined for the run created notification:

ibutsu-import

3) Find the related run (use the developer tools on the network tab to find the run ID) and check the API GET response and verify that the `project_id` is `null`:

```json
{
  "component": null,
  "created": "2020-12-08T14:09:33.739307Z",
  "data": null,
  "duration": 0.067,
  "env": null,
  "id": "c3d3ea87-ca0b-4255-8d14-510e6d2cb0b1",
  "project_id": null,
  "source": null,
  "start_time": "2020-11-17T14:51:48.114645Z",
  "summary": {
    "errors": "1",
    "failures": "1",
    "skips": "1",
    "tests": "4",
    "xfailures": null,
    "xpasses": null
  }
}
```

Another suggestion would be to allow the user to close the run created notification so they would have time to click if it provides the run link. Or increase its timeout.

rsnyman:
Fix up some paper cuts

### DIFF
--- a/frontend/src/app.js
+++ b/frontend/src/app.js
@@ -276,6 +276,10 @@ class App extends React.Component {
         </NavList>
       </Nav>
     );
+    const uploadParams = {};
+    if (this.state.selectedProject && this.state.selectedProject.project) {
+      uploadParams['project'] = this.state.selectedProject.project.id;
+    }
     const topNav = (
       <Flex>
         <FlexItem id="project-selector">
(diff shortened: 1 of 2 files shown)


john-dupuy (approved):
LGTM, testing this locally with Ibutsu's unit tests, it works well.

PR #118, opened 2020-12-11
Update API spec to allow null values in Runs

### LINKED ISSUE #111: Run GET response can't be used directly into a PUT request
Considering the following run GET response (one provided after a jUnit import):

```json
{
  "component": null,
  "created": "2020-12-08T14:09:33.739307Z",
  "data": null,
  "duration": 0.067,
  "env": null,
  "id": "c3d3ea87-ca0b-4255-8d14-510e6d2cb0b1",
  "project_id": null,
  "source": null,
  "start_time": "2020-11-17T14:51:48.114645Z",
  "summary": {
    "errors": "1",
    "failures": "1",
    "skips": "1",
    "tests": "4",
    "xfailures": null,
    "xpasses": null
  }
}
```

Then when trying to use it on a PUT the following 400 response body is returned:

```json
{
  "detail": "None is not of type 'string' - 'source'",
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank"
}
```

So it seems that source is expected to be an empty string `""` instead of null but the GET response for a run is not providing a valid value. If I just drop the `source` field it works as expected.

rsnyman:
Closes #111

### DIFF
--- a/backend/ibutsu_server/openapi/openapi.yaml
+++ b/backend/ibutsu_server/openapi/openapi.yaml
@@ -1503,16 +1503,33 @@ components:
           description: A source for this test run
           example: my-tests
           type: string
+          nullable: true
         start_time:
           description: The time the test run started
           example: 2020-05-15T16:18:32.014053
           type: string
+        component:
+          description: A component
+          example: login
+          type: string
+          nullable: true
+        env:
+          description: The environment which is being tested
+          example: qa
+          type: string
+          nullable: true
+        project_id:
+          description: The project this run is associated with
+          example: 44941c55-9736-42f6-acce-ca3c4739d0f3
+          type: string
+          nullable: true
         summary:
           description: A summary of the test results
           type: object
         metadata:
-          description: Extra data for this run
+          description: Extra metadata for this run
           type: object
+          nullable: true
       type: object
     UpdateRun:
       example:


backend/ibutsu_server/openapi/openapi.yaml
```diff
@@ -1483,7 +1483,7 @@ components:
           xfailures: 0
           xpasses: 1
           tests: 548
-        metadata:
+        data:
```
john-dupuy, line 1486: Though this is called `data` in the DB model, in Ibutsu's API and FE we still refer to it as `metadata`, and do the translation to `data` in the `ModelMixin`

https://github.com/ibutsu/ibutsu-server/blob/721a77db65f78d651effb0db35c53c487c5c9378/backend/ibutsu_server/db/models.py#L44-L45

```suggestion,
        metadata:
```

backend/ibutsu_server/openapi/openapi.yaml
```diff
@@ -1503,16 +1503,33 @@ components:
...
+        project_id:
+          description: The project this run is associated with
+          example: 44941c55-9736-42f6-acce-ca3c4739d0f3
+          type: string
+          nullable: true
         summary:
           description: A summary of the test results
           type: object
-        metadata:
+        data:
```
john-dupuy, line 1529: Same comment about `data/metadata` here:
```suggestion
        metadata:
```

backend/ibutsu_server/openapi/openapi.yaml
rsnyman: Ah, OK. I thought so, except the example said, "data" and not "metadata"

john-dupuy (approved):

PR #131, opened 2021-01-05
Add a upgrade function to add gin indexes on metadata.tags and metadata.requirements

john-dupuy:
Adding an `upgrade_2` function to add indexes to aid in queries that search `metadata.tags` and `metadata.requirements`

### DIFF
--- a/backend/ibutsu_server/db/upgrades.py
+++ b/backend/ibutsu_server/db/upgrades.py
@@ -56,11 +56,14 @@ def upgrade_1(session):
 def upgrade_2(session):
     """Version 2 upgrade
 
-    This upgrade adds indices for the metadata.tags fields in the results/runs tables.
+    This upgrade adds indices for the metadata.tags, metadata.requirements fields in the
+    results/runs tables.
 
     It is equivalent to running the following SQL statements:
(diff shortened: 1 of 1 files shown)


backend/ibutsu_server/db/upgrades.py
```diff
@@ -50,3 +51,24 @@ def upgrade_1(session):
...
+    This upgrade adds indices for the metadata.tags fields in the results/runs tables.
+
+    It is equivalent to running the following SQL statements:
+        CREATE INDEX IF NOT EXISTS ix_runs_tags ON runs USING gin ((data->'tags'));
+        CREATE INDEX IF NOT EXISTS ix_results_tags ON results USING gin ((data->'tags'));
+    """
+    TABLES = ["runs", "results"]
+
+    op = get_upgrade_op(session)
+    for table in TABLES:
```
rsnyman, line 68: I think we probably want to just check that the engine is postgres before running this. Just in case someone decides to run this locally using SQLite or something.

rsnyman (changes requested):

backend/ibutsu_server/db/upgrades.py
john-dupuy: Ah, good idea, updated to check the database type.

rsnyman (approved):

PR #133, opened 2021-01-06
Added EmptyObject component for missing results/runs

jjaquish:


frontend/src/components/empty-object.js
```diff
@@ -0,0 +1,50 @@
...
+    returnLinkText: PropTypes.string
+  };
+  render() {
+    return (
+      <React.Fragment>
+        <EmptyState>
+          <EmptyStateIcon icon={SearchIcon} />
+          <Text headingLevel="h1" size="lg">
+            {!this.props.headingText &&
+            <Text>This object couldn&apos;t be found.</Text>}
```
john-dupuy, line 31:
```suggestion
            {!this.props.headingText &&
            <Text>This object couldn&apos;t be found.</Text>
            }
```

frontend/src/components/empty-object.js
```diff
@@ -0,0 +1,50 @@
...
+            {this.props.headingText}
```
john-dupuy, line 32:
```suggestion
            {this.props.headingText ? this.props.headingText : "This object couldn't be found."}
```

frontend/src/components/empty-object.js
```diff
@@ -0,0 +1,50 @@
...
+          </Text>
+          <EmptyStateBody>
+            {!this.props.bodyText &&
+            <Text>Either the object doesn&apos;t exist or the ID is invalid.</Text>}
+            {this.props.bodyText}
```
john-dupuy, line 37: Consider a similar ternary as the one suggested above.

frontend/src/components/empty-object.js
```diff
@@ -0,0 +1,50 @@
...
+          </EmptyStateBody>
+          <NavLink style={{ color: 'white' }} to={!this.props.returnLink ?  '' : this.props.returnLink}>
+            <Button variant="primary" style = {{ margin: '25px' }}>
+              {!this.props.returnLinkText &&
+              <Text>Return to dashboard</Text>}
+              {this.props.returnLinkText}
```
john-dupuy, line 43: Consider a similar ternary as the one above.

frontend/src/result.js
```diff
@@ -32,7 +32,13 @@ export class Result extends React.Component {
...
     }
     fetch(Settings.serverUrl + '/result/' + this.state.id)
-      .then(response => response.json())
+      .then(response => {
+        this.setState({ "httpStatus": response.status });
+        if (!response.ok) {
+          throw new Error("Failed with HTTP code " + response.status);
+        }
+        return response.json();
+      })
```
john-dupuy, line 41: Rather than having the `httpStatus` be a state variable, I think we should make `isResultValid` (or `isNotFound`) a state variable and update it here.

That way we can avoid the conditional in the `render` function below.

frontend/src/run.js
```diff
@@ -341,10 +343,17 @@ export class Run extends React.Component {
...
   getRun() {
     fetch(Settings.serverUrl + '/run/' + this.state.id)
-      .then(response => response.json())
+      .then(response => {
+        this.setState({ "httpStatus": response.status });
+        if (!response.ok) {
+          throw new Error("Failed with HTTP code " + response.status);
+        }
+        return response.json();
+      })
```
john-dupuy, line 352: Same comment here about `httpStatus` and the `isNotFound`/`isRunValid` state variable.

john-dupuy (changes requested):
Nice work @jjaquish! I have a few suggested changes.

frontend/src/result.js
```diff
@@ -49,11 +57,16 @@ export class Result extends React.Component {
       <React.Fragment>
         <PageSection variant={PageSectionVariants.light}>
           <TextContent>
-            <Text component="h1">{testResult && testResult.test_id}</Text>
+            <Text component="h1">
+              {testResult && testResult.test_id}
+              {!testResult && <Text>Result</Text>}
```
john-dupuy, line 62:
```suggestion
              {testResult ?  testResult.test_id : <Text>Result</Text>}
```

frontend/src/result.js
```diff
@@ -49,11 +57,16 @@ export class Result extends React.Component {
...
+            </Text>
           </TextContent>
         </PageSection>
         <PageSection>
-          <ResultView testResult={testResult} history={this.props.history} location={this.props.location}/>
+          {!this.state.isResultValid &&
+          <EmptyObject headingText="Result not found" returnLink="/results" returnLinkText="Return to results list" />}
+          {this.state.isResultValid && <ResultView testResult={testResult} history={this.props.history} location={this.props.location}/>}
```
john-dupuy, line 69:
```suggestion
          {this.state.isResultValid ? 
           <ResultView testResult={testResult} history={this.props.history} location={this.props.location}/> :
           <EmptyObject .../>
          }
```

john-dupuy (changes requested):
Two small suggestions about using more `ternary`

PR #140, opened 2021-01-29
Fix filters on reports

john-dupuy:
Thanks @mshriver for reporting the issue! 

Fixes this exception that shows up whenever reports are filtered:
```
  File "/app/ibutsu_server/tasks/reports.py", line 132, in _get_results
    query = _build_query(report)
  File "/app/ibutsu_server/tasks/reports.py", line 120, in _build_query
    filter_clause = convert_filter(Result, report_filter.strip())
  File "/app/ibutsu_server/filters.py", line 50, in convert_filter
    match = FILTER_RE.match(filter_string)
TypeError: expected string or bytes-like object
```

### DIFF
--- a/backend/ibutsu_server/tasks/reports.py
+++ b/backend/ibutsu_server/tasks/reports.py
@@ -10,7 +10,7 @@ from ibutsu_server.db.base import session
 from ibutsu_server.db.models import Report
 from ibutsu_server.db.models import ReportFile
 from ibutsu_server.db.models import Result
-from ibutsu_server.filters import convert_filter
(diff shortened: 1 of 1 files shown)


mshriver (approved):
Take my review with a grain of salt, but `ibutsu_server.filters.apply_filters` does indeed call `convert_filter` with the filter string as the first positional argument and the `Result` model as the second positional argument.  In general it certainly appears to be a better practice to use `apply_filters` here instead of implementing the same statements locally to `_build_query`

rsnyman (approved):

PR #141, opened 2021-02-11
Add in tasks to prune old runs and results from the DB

john-dupuy:
* Prune runs, results on weekly basis
* Runs older than 12 months will be removed 
* Results older than 6 months will be removed
* Fixing a bug that was deleting artifact files prematurely

### DIFF
--- a/backend/ibutsu_server/tasks/db.py
+++ b/backend/ibutsu_server/tasks/db.py
@@ -4,23 +4,79 @@ from datetime import timedelta
 
 from ibutsu_server.db.base import session
 from ibutsu_server.db.models import Artifact
+from ibutsu_server.db.models import Result
+from ibutsu_server.db.models import Run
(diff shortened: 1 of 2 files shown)


backend/ibutsu_server/tasks/db.py
```diff
@@ -4,25 +4,81 @@
...
     try:
         if isinstance(months, str):
             months = int(months)
 
         if months < 2:
-            # we don't want to remove files more recent than 3 months
+            # we don't want to remove files more recent than 2 months
             return
 
-        max_date = datetime.utcnow() - timedelta(days=months * 4)
```
john-dupuy, line 21: Oops :grimacing:

rsnyman: :laughing:

rsnyman (approved):

PR #147, opened 2021-02-17
Output number of collected tests

john-dupuy:
* Remove some of the duplicated logic in the pytest-plugin
* Use the number of collected tests in the frontend

Requires https://github.com/ibutsu/pytest-ibutsu/pull/17

@rsnyman Will this break the imports? I'm debating whether we can get rid of the loop over the results in the update_run task altogether, all we need is 1 result to copy the metadata/columns over.

Note the pass percentages are still determined from `run.summary.tests`. Meant to address https://github.com/ibutsu/pytest-ibutsu/issues/13

Screenshot from 2021-02-17 14-49-20

### DIFF
--- a/backend/ibutsu_server/tasks/importers.py
+++ b/backend/ibutsu_server/tasks/importers.py
@@ -97,7 +97,9 @@ def run_junit_import(import_):
         import_record.data["run_id"].append(run.id)
         # Import the contents of the XML file
         for testcase in testsuite.iterchildren(tag="testcase"):
-            test_name = testcase.get("name").split(".")[-1]
+            test_name = ""
+            if testcase.get("name"):
+                test_name = testcase.get("name").split(".")[-1]
             backup_fspath = None
             if testcase.get("classname"):
                 test_name = testcase.get("classname").split(".")[-1] + "." + test_name
--- a/backend/ibutsu_server/tasks/runs.py
+++ b/backend/ibutsu_server/tasks/runs.py
@@ -23,40 +23,42 @@ def _copy_column(result, run, key):
 def update_run(run_id):
     """Update the run summary from the results, this task will retry 1000 times"""
     with lock(f"update-run-lock-{run_id}"):
-        key_map = {
-            "failed": "failures",
-            "error": "errors",
-            "skipped": "skips",
-            "xpassed": "xpasses",
-            "xfailed": "xfailures",
-        }
         run = Run.query.get(run_id)
         if not run:
             return
-        # sort according to starttime to get the most recent starting time of the run
+        # sort according to start_time to get the most recent starting time of the run
         results = (
-            Result.query.filter(Result.run_id == run_id).order_by(Result.start_time.asc()).all()
+            Result.query.filter(Result.run_id == run_id).order_by(Result.start_time.asc()).limit(1)
         )
-        summary = {"errors": 0, "failures": 0, "skips": 0, "tests": 0, "xpasses": 0, "xfailures": 0}
-        run_duration = 0.0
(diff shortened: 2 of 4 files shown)


backend/ibutsu_server/tasks/importers.py
```diff
@@ -97,7 +97,9 @@ def run_junit_import(import_):
         import_record.data["run_id"].append(run.id)
         # Import the contents of the XML file
         for testcase in testsuite.iterchildren(tag="testcase"):
-            test_name = testcase.get("name").split(".")[-1]
+            test_name = None
```
rsnyman, line 100: Maybe set this to a blank string? Otherwise line 105 below is going to börk.

backend/ibutsu_server/tasks/runs.py
```diff
@@ -23,38 +23,35 @@ def _copy_column(result, run, key):
...
-        run_duration = 0.0
+        summary = {
+            "errors": run.summary.get("errors", 0),
+            "failures": run.summary.get("failures", 0),
+            "skips": run.summary.get("skips", 0),
+            "tests": run.summary.get("tests", 0),
+            "xpasses": run.summary.get("xpasses", 0),
+            "xfailures": run.summary.get("xfailures", 0),
+            "tests_collected": run.summary.get("tests_collected", 0),
+        }
```
rsnyman, line 41: Since you're changing this, can you add in "passes" too? I'm tired of calculating the passes on the fly every time.

frontend/src/run.js
```diff
@@ -460,6 +460,10 @@ export class Run extends React.Component {
         passed -= run.summary.skips;
```
rsnyman, line 460: Once you've changed the backend to put the passed results into a "passes" field, you can update this to use that field if it exists.

backend/ibutsu_server/tasks/runs.py
```diff
@@ -23,38 +23,35 @@ def _copy_column(result, run, key):
...
-        summary = {"errors": 0, "failures": 0, "skips": 0, "tests": 0, "xpasses": 0, "xfailures": 0}
-        run_duration = 0.0
+        summary = {
+            "errors": run.summary.get("errors", 0),
+            "failures": run.summary.get("failures", 0),
+            "skips": run.summary.get("skips", 0),
+            "tests": run.summary.get("tests", 0),
+            "xpasses": run.summary.get("xpasses", 0),
+            "xfailures": run.summary.get("xfailures", 0),
+            "tests_collected": run.summary.get("tests_collected", 0),
```
rsnyman, line 40: Can we maybe just call this "collected"?

rsnyman: I realise I just merged the other PR that uses this field. I think I'm going to submit a PR to fix that one.

rsnyman (changes requested):

backend/ibutsu_server/tasks/runs.py
rsnyman: https://github.com/ibutsu/pytest-ibutsu/pull/18

backend/ibutsu_server/tasks/runs.py
john-dupuy: @rsnyman we would need to calculate the `passes` in the pytest_plugin in order to do that. I can open a PR for that if you'd like.

rsnyman: I think we should do it in the plugin too. Doing it here will make it backward compatible, which is not a bad thing :-)

rsnyman (approved):

PR #154, opened 2021-03-05
Use 'id' rather than 'name' in dashboard GET

john-dupuy:
Thanks @mshriver for reporting this one :smiley:


mshriver:
I just can't use anything without breaking it a bit.

rsnyman (approved):

Issue #167, opened 2021-05-17
Update runs that have not been gracefully exitted


john-dupuy:
When a pytest session is interrupted, e.g. by an outage or by the pod being killed, the run is never updated. This means that the counts, metadata, etc will never be added to the run. 

We should have a periodic task that goes through and checks for these runs - updating them with the partial results.

PR #168, opened 2021-05-17
Update pre-commit; add in task for syncing aborted runs

john-dupuy:
Address #167

* Update pre-commit
* Add in a periodic task (runs every 15 minutes) to look at recent runs and sync `summary.tests` with the actual number of results

Change the way the `update_run` task works (basically reverting some of the changes made in #147), although this has redundancy with the pytest-ibutsu plugin (summary is being computed again), this is the cleanest way to make sure the runs are synced with the results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,6 @@ backend/settings.yaml
 # pytest-ibutsu stuff
 *.tar.gz
 .last*
+
+# celery beat
+celerybeat*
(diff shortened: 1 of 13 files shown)


rsnyman:
I see that flask_testing is pinned. If I remove that pin, the Werkzeug pin and upgrade all three (Flask, flask_testing and Werkzeug), then the tests run fine.

john-dupuy:
Aha! Thanks for looking into this, let me try that :+1:

rsnyman (approved):
Looks good to me.

Issue #163, opened 2021-04-07
Run large queries in tasks


rsnyman:
Large results from queries to the API can timeout and crash. It may be a good idea to look into a method of running queries in tasks.

Steps:
1. Client submits a query to the API
2. API plans the query and determines a large amount of rows will be queried/returned
3. API returns 201/202 and a query ID
4. Query is run in a task
5. Client polls API with query ID
6. On completion, the API returns the results

PR #174, opened 2021-05-24
Perform large queries for results/runs in a task

john-dupuy:
* Add task API endpoint 
* Perform large queries in a task
* Return a 201 if page_size is above some limit

This work is for #163


rsnyman (commented):
Dude. THIS. IS. MAGIC.

Issue #192, opened 2021-06-07
Allow some env vars to be passed in at runtime for the frontend


john-dupuy:
Currently, we have https://github.com/ibutsu/ibutsu-server/blob/master/frontend/src/settings.js which contains some basic settings for the application.

In OCP, we have to specify these settings (in particular `REACT_APP_SERVER_URL`) at build time of the container image rather than at run time. It would be nice if we could specify them at run time. 

Having this `settings.js` file creates a mismatch between the env vars in the container image and the env var of the runtime container. 

For instance, if you
- Build the frontend image
- Run it like `docker run -e NODE_ENV=development -p "3000:8080" ibutsu/frontend`

`NODE_ENV` of the runtime container will be `development` but `process.env.NODE_ENV` in the `settings.js` file resolves to `production`, because that was the env var of the image.

john-dupuy:
https://github.com/joellord/frontend-containers/blob/main/envvars.md

john-dupuy:
Thinking about this some more, it may be better to simply build the production frontend as a part of a GH action and push it up to quay on release. That way we can just specify the build environment variable (afaik this cannot be done in quay). Then we can have production/local images.

PR #188, opened 2021-06-02
Small changes to dockerfiles

john-dupuy:
* Update flower/monitor with proper changes for the new image
* Change the frontend and backend container images to the appropriate s2i scripts
  - https://github.com/sclorg/s2i-python-container/tree/master/3.8 for the backend
  - https://github.com/sclorg/s2i-nodejs-container/tree/master/14 for the frontend
* Remove root `Dockerfile` that is no longer used
* Fix/update docker-compose (I am not very familiar with docker-compose but I wasn't able to get it working without `network_mode: host`), anyways, I think https://github.com/ibutsu/ibutsu-server/pull/191 using `podman pods` is a much better strategy

The changes to the dockerfiles for the frontend/backend are meant to mimic the s2i strategy we use in OCP, for more information, see the README's at the `sclorg` links above.

PR #193, opened 2021-06-08
Frontend env vars at container runtime

john-dupuy:
Address https://github.com/ibutsu/ibutsu-server/issues/192 by:
- creating a `public/settings.js` file that is loaded before react
- adding a `start-frontend.js` that overwrites the `public/settings.js` file at container runtime
- changing `src/settings.js` to be assigned to window.settings

Depends on #188

### DIFF
--- a/frontend/docker/Dockerfile.frontend
+++ b/frontend/docker/Dockerfile.frontend
@@ -5,6 +5,9 @@ FROM registry.access.redhat.com/ubi8/nodejs-14
 # and set permissions so that the container runs without root access
 USER 0
 ADD . /tmp/src
+COPY ./docker/start-frontend.sh /usr/bin/start-frontend.sh
+RUN chmod +x /usr/bin/start-frontend.sh
+RUN chown 1001:0 /usr/bin/start-frontend.sh
 RUN chown -R 1001:0 /tmp/src
 USER 1001
 
@@ -12,4 +15,4 @@ USER 1001
 RUN /usr/libexec/s2i/assemble
 
 # Set the default command for the resulting image
-CMD /usr/libexec/s2i/run
+ENTRYPOINT [ "start-frontend.sh" ]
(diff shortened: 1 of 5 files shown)


rsnyman:
@john-dupuy what do you think of making an abstraction layer between the window.settings object and the code? So rather than using the object directly, use the Settings object as a wrapper/interim. It also looks cleaner :-)

```javascript
// public/settings.js
window.settings = {...};
```

```javascript
// src/settings.js
Settings = window.settings;
```

```javascript
// other code
import { Settings } from './settings';

fetch(Settings.serverUrl)
```

john-dupuy:
> @john-dupuy what do you think of making an abstraction layer between the window.settings object and the code? So rather than using the object directly, use the Settings object as a wrapper/interim. It also looks cleaner :-)
> 
> ```js
> // public/settings.js
> window.settings = {...};
> ```
> 
> ```js
> // src/settings.js
> Settings = window.settings;
> ```
> 
> ```js
> // other code
> import { Settings } from './settings';
> 
> fetch(Settings.serverUrl)
> ```
Ah! That's basically what I had here: https://github.com/ibutsu/ibutsu-server/pull/193/commits/1b5258b1c33b482246b51ed2c97c9d501aa3c119. I must've misunderstood your request earlier. I agree it looks cleaner :) - will update.

rsnyman:
> Ah! That's basically what I had here (minus the more apt names): [1b5258b](https://github.com/ibutsu/ibutsu-server/commit/1b5258b1c33b482246b51ed2c97c9d501aa3c119). I must've misunderstood your request earlier. I agree it looks cleaner :) - will update.

Ja, sorry, I was only half-awake when I look at it earlier ;-)

john-dupuy:
Haha no worries - just updated keeping `Settings` in src/settings.js

rsnyman (approved):

PR #198, opened 2021-06-15
Update OCP template config

john-dupuy:
* get rid of task-monitor, with the `task` api endpoint, we can use that to check the status of tasks
* remove all build configs now that builds are handled in quay
* rename image-streams from `ibutsu-<app>` to just `<app>`
* update image-streams to point to images on quay
* add env vars to frontend now that #193 is deployed


ocp-templates/ibutsu-server-openshift.yaml
```diff
@@ -66,7 +26,12 @@ objects:
           deploymentConfig: ibutsu-frontend
       spec:
         containers:
-        - image: ibutsu-frontend
+        - env:
+            - name: REACT_APP_SERVER_URL
+              value: ${BACKEND_ROUTE}
+            - name: NODE_ENV
+              value: production
+        - image: frontend
```
john-dupuy, line 34: I'm not 100% sure if this is necessary or if it needs to be `quay.io/ibutsu/frontend`?

ocp-templates/ibutsu-server-openshift.yaml
```diff
@@ -104,21 +69,31 @@ objects:
...
-      local: true
+      local: false
+    tags:
+    - annotations: null
+      from:
+        kind: DockerImage
+        name: quay.io/ibutsu/frontend
+      generation: 3
+      importPolicy:
+        scheduled: true
```
john-dupuy, line 93: Note: this will check for new images on quay every 15 minutes. So when releasing, a new deployment may not happen until after ~15 mins of a deploy on quay.

ocp-templates/ibutsu-server-openshift.yaml
rsnyman: Hrm, yes, I think it needs to be `quay.io/ibutsu/frontend` -- try it and see if there are any problems :-)

ocp-templates/ibutsu-server-openshift.yaml
rsnyman: I think 15 minutes is OK.

ocp-templates/ibutsu-server-openshift.yaml
john-dupuy: Looking into this a bit more, it seems like for native OCP objects (DeploymentConfig's included) it is enough to just use the image stream name as we are here. OCP will resolve the image from the image stream name.

For Deployments, if we wanted to enable this we would have to change the `spec.lookupPolicy.local` to `true`. 

https://docs.openshift.com/container-platform/4.6/openshift_images/using-imagestreams-with-kube-resources.html
https://austindewey.com/2020/10/25/you-can-use-openshifts-imagestreams-with-deployments-heres-how/

rsnyman (approved):

Issue #205, opened 2021-09-14
Use the tasks endpoint to check for failed reports


john-dupuy:
Right now, failed reports just show as "running". We should use the `tasks` endpoint to mark them as failed/error/etc.

PR #209, opened 2021-09-15
Set failed reports as error

john-dupuy:
#205 

I didn't end up needing the tasks endpoint, but this is a way to set those reports as error without having to externally check them.

* also remove the `monitor.py` file that is no longer used


jjaquish (commented):
LGTM initially, but I didn't test this.

rsnyman (approved):

PR #215, opened 2021-10-01
Artifacts on Runs

john-dupuy:


rsnyman (approved):
This looks good. The only change I might investigate is making a more generic `getArtifacts()` method that maybe takes a set of parameters as an argument, so that we can use the same logic in more than one place without duplicating the code.

For example:

```javascript
getArtifacts(params) {
  fetch(buildUrl(Settings.serverUrl + '/artifact', params))
  ...
}
```

john-dupuy:
Oh yeah that's a good idea, I can move it out to the utilities function. I'll work on that today.

backend/ibutsu_server/db/upgrades.py
```diff
@@ -33,7 +33,7 @@ def upgrade_1(session):
     if (
         "dashboards" in metadata.tables
         and widget_configs is not None
-        and not widget_configs.columns.contains_column("dashboard_id")
+        and not hasattr(widget_configs.columns, "dashboard_id")
```
john-dupuy, line 36: Try `contains_column` with the `column` object instead? 

https://github.com/sqlalchemy/sqlalchemy/blob/a162c1d714fe16fefb5da11981a4e788ca195d6a/lib/sqlalchemy/sql/base.py#L1320-L1323

john-dupuy: Turns out you can do
```python
column = widget_configs.columns.get("dasbhoard_id")
widget_configs.columns.contains_column(column)
``` 
And that will return true. However, this is more verbose than just using `hasattr`, so I'm going to leave it.

PR #221, opened 2021-10-12
Add MetaFilter component, use on ClassifyFailures

mshriver:
Add a metadata filtering component, and tie it to ClassifyFailures
FilterTable

First time working with React, not at all sure if I did this 'right'.

At this point I've tested locally, but not extensively.  Opening this for some input from @rsnyman @john-dupuy @jjaquish  on my approach here.

There's also been discussion of doing this more generically, like the run/result filter page, with separate dropdowns for field and value selection.

## Preview
image

### DIFF
--- a/frontend/src/components/classify-failures.js
+++ b/frontend/src/components/classify-failures.js
@@ -9,9 +9,6 @@ import {
   Checkbox,
   Flex,
   FlexItem,
-  Select,
-  SelectOption,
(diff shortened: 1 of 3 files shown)


john-dupuy:
> Not sure why lint is failing, pre-commit runs cleanly for me.
Investigating now, definitely unrelated to your changes. The same thing happened in #220

PR #225, opened 2021-10-12
Implement MetaFilter with separate field and value

mshriver:
Alternative to #221 that is more generic. Similar to the filters used on
the run/results page, and could be an abstract replacement for their
current filters.

## TODO
- [x] complete functions for field/value clear
- [x] complete function for value select
- [x] add checkbox to apply `not` operator (`eq` and `in` will be inferred)  #230 
- [x] handling of null value metadata (key only) :: Specific list of metadata fields that can be filtered has been added. Outside of this PR scope to further abstract this.
- [x] handling of non-string metadata, dates, etc. :: Specific list of metadata fields that can be filtered has been added. Outside of this PR scope to further abstract this.
- [ ] Rebase to account for Auth

### Nothing selected
image

### Field selected with no values available
image

### Field selected with values available/selected
image

### DIFF
--- a/frontend/src/components/classify-failures.js
+++ b/frontend/src/components/classify-failures.js
@@ -9,9 +9,6 @@ import {
   Checkbox,
   Flex,
   FlexItem,
-  Select,
-  SelectOption,
-  SelectVariant,
   TextContent,
   Text,
 } from '@patternfly/react-core';
@@ -24,13 +21,15 @@ import { HttpClient } from '../services/http';
 import { Settings } from '../settings';
 import {
   buildParams,
+  toAPIFilter,
   getSpinnerRow,
   resultToClassificationRow,
 } from '../utilities';
-import { OPERATIONS } from '../constants';
+import { FILTERABLE_RESULT_FIELDS } from '../constants';
 import {
   FilterTable,
   MultiClassificationDropdown,
+  MetaFilter,
 } from './index';
 
 
@@ -38,7 +37,7 @@ export class ClassifyFailuresTable extends React.Component {
   static propTypes = {
     filters: PropTypes.object,
     run_id: PropTypes.string
-  };
+  }
 
   constructor(props) {
     super(props);
@@ -56,9 +55,6 @@ export class ClassifyFailuresTable extends React.Component {
       isError: false,
       isFieldOpen: false,
       isOperationOpen: false,
-      exceptionSelections: [],
-      isExceptionOpen: false,
-      exceptions: [],
       includeSkipped: false,
       filters: Object.assign({
         'result': {op: 'in', val: 'failed;error'},
@@ -72,39 +68,6 @@ export class ClassifyFailuresTable extends React.Component {
     this.setState({selectedResults: []});
     this.getResultsForTable();
   }
-
-  onExceptionToggle = isExpanded => {
-    this.setState({isExceptionOpen: isExpanded}, this.applyFilter);
-  }
-
-  applyExceptionFilter = () => {
-    let { filters, exceptionSelections } = this.state;
-    if (exceptionSelections.length > 0) {
-      filters["metadata.exception_name"] = Object.assign({op: 'in', val: exceptionSelections.join(';')});
-      this.setState({filters}, this.refreshResults);
-    }
-    else {
-      delete filters["metadata.exception_name"];
-      this.setState({filters}, this.refreshResults);
-    }
-  }
-
-  onExceptionSelect = (event, selection) => {
-    const exceptionSelections = this.state.exceptionSelections;
-    if (exceptionSelections.includes(selection)) {
-      this.setState({exceptionSelections: exceptionSelections.filter(item => item !== selection)}, this.applyExceptionFilter);
-    }
-    else {
-     this.setState({exceptionSelections: [...exceptionSelections, selection]}, this.applyExceptionFilter);
-    }
-  };
-
-  onExceptionClear = () => {
-    this.setState({
-      exceptionSelections: [],
-      isExceptionOpen: false
-    }, this.applyExceptionFilter);
-  };
 
   onCollapse(event, rowIndex, isOpen) {
     const { rows } = this.state;
@@ -146,7 +109,7 @@ export class ClassifyFailuresTable extends React.Component {
 
   updateFilters(name, operator, value, callback) {
     let filters = this.state.filters;
-    if (!value) {
+    if ((value === null) || (value.length === 0)) {
       delete filters[name];
     }
     else {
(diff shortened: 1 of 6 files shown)


frontend/src/components/classify-failures.js
```diff
@@ -73,39 +70,18 @@ export class ClassifyFailuresTable extends React.Component {
...
 
-  onExceptionToggle = isExpanded => {
-    this.setState({isExceptionOpen: isExpanded}, this.applyFilter);
-  }
-
-  applyExceptionFilter = () => {
-    let { filters, exceptionSelections } = this.state;
-    if (exceptionSelections.length > 0) {
-      filters["metadata.exception_name"] = Object.assign({op: 'in', val: exceptionSelections.join(';')});
+  applyFilter = (field, value) => {
```
mshriver, line 73: I'm looking at updating the existing updateFilters function on this class to use instead of this.

frontend/src/components/classify-failures.js
```diff
@@ -248,35 +199,25 @@ export class ClassifyFailuresTable extends React.Component {
...
-            <SelectOption key={index} value={option._id} description={option.count + ' results'}/>
-          ))}
-        </Select>
-      </React.Fragment>
+    // filters for the metadata
+    const resultFilters = [
+      <MetaFilter
+        key="metafilter"
+        // user_properties fields shouldn't be injected here
+        fieldOptions={['metadata.user_properties.assignee', 'metadata.user_properties.component', ...STRING_RESULT_FIELDS,]}
```
john-dupuy, line 216: There are some `fieldOptions` like `run_id` and `project_id` which are included in `STRING_RESULT_FIELDS` that aren't really relevant here. 

I think we should just leave with a few fields for now. Also, I noticed `metadata.exception_name` isn't included in the list.

frontend/src/components/classify-failures.js
```diff
@@ -248,35 +199,25 @@ export class ClassifyFailuresTable extends React.Component {
...
+        runId={run_id}
+        setFilter={this.setFilter}
+        customFilters={['result*failed;error']}
```
john-dupuy, line 219: Does the include `skips/xfails` checkbox factor in here?

mshriver:
Thanks @john-dupuy! I should be able to work through those issues today.

frontend/src/components/classify-failures.js
mshriver: One piece that I _might_ leave as a TODO here is more abstract handling of adding fields to this list.

Similar to the values select, we can try to resolve a set of fields from the result items metadata - to include things like user_properties without having to enumerate beforehand.

frontend/src/components/filtertable.js
```diff
@@ -160,3 +168,125 @@ export class FilterTable extends React.Component {
...
+      })
+    }
+  }
+
+  render () {
+    const {isFieldOpen, fieldSelection, isValueOpen, valueOptions, valueSelections} = this.state
+
+    return (
+      <React.Fragment>
+        <Select key="metafield_select"
```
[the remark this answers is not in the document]
mshriver: I'll let others weigh in. I don't have a strong opinion here, but my inclination is to not have a default.

For me and Jake exception_name would be the preferred default. For everyone else on the team it'll be assignee or component - because they want to see their own failures, not everybody else's.

frontend/src/components/classify-failures.js
mshriver: It does now!  Validated with some results that had skips and the checkbox works as expected now and impacts the counts of value select options.

frontend/src/components/filtertable.js
```diff
@@ -160,3 +169,135 @@ export class FilterTable extends React.Component {
...
+    let field_selected = this.state.fieldSelection !== null
+    let values_available = valueOptions.length > 0
+    let value_placeholder = "Select a field first"  // default instead of an else block
+    if (field_selected && values_available){ value_placeholder = "Select value(s)"}
+    else if (field_selected && !values_available) { value_placeholder = "No Values for selected field"}
+    return (
+      <React.Fragment>
+        <Select key="metafield_select"
+          aria-label="metadata-field-filter"
+          placeholderText="Select Metadata Field"
```
john-dupuy, line 270: To match casing of the other select:
```suggestion
          placeholderText="Select metadata field"
```

frontend/src/components/filtertable.js
```diff
@@ -160,3 +169,135 @@ export class FilterTable extends React.Component {
...
+    }
+  }
+
+  render () {
+    const {isFieldOpen, fieldSelection, isValueOpen, valueOptions, valueSelections} = this.state
+    let field_selected = this.state.fieldSelection !== null
+    let values_available = valueOptions.length > 0
+    let value_placeholder = "Select a field first"  // default instead of an else block
+    if (field_selected && values_available){ value_placeholder = "Select value(s)"}
+    else if (field_selected && !values_available) { value_placeholder = "No Values for selected field"}
```
john-dupuy, line 265:
```suggestion
    else if (field_selected && !values_available) { value_placeholder = "No values for selected field"}
```

frontend/src/components/classify-failures.js
```diff
@@ -156,27 +119,20 @@ export class ClassifyFailuresTable extends React.Component {
...
   removeFilter = id => {
-    if (id === "metadata.exception_name") {   // only remove exception_name filter
-      this.updateFilters(id, null, null, () => {
-        this.setState({exceptionSelections: [], page: 1}, this.applyExceptionFilter);
-      });
+    if ((id !== "result") && (id !== "run_id")) {   // Don't allow removal of error/failure filter
+      this.updateFilters(id, null, null, this.refreshResults)
     }
-  }
+  };
```
rsnyman, line 131: I know we're inconsistent here, but the semi-colon is technically optional. Don't worry about it in the future :smile:

frontend/src/constants.js
```diff
@@ -60,6 +60,26 @@ export const STRING_RESULT_FIELDS = [
...
+  // TODO support array filtering
+  //'metadata.statuses.call',
+  //'metadata.statuses.setup',
+  //'metadata.statuses.teardown',
+  'metadata.caseautomation',
+  'metadata.component',
+  'metadata.exception_name',
+  'metadata.fspath',
+  'metadata.importance',
+  'metadata.project',
```
john-dupuy, line 76: I think we can remove this one.

frontend/src/constants.js
```diff
@@ -60,6 +60,26 @@ export const STRING_RESULT_FIELDS = [
...
+export const FILTERABLE_RESULT_FIELDS = [
+  'env',
+  'component',
+  'metadata.assignee',
+  // TODO support array filtering
+  //'metadata.statuses.call',
+  //'metadata.statuses.setup',
+  //'metadata.statuses.teardown',
+  'metadata.caseautomation',
+  'metadata.component',
```
john-dupuy, line 72: I think we can remove this one too, it's covered by just `component`.

mshriver:
@john-dupuy thanks for your diligence in testing - I think the value count behavior tied to skip/xfail is the same root cause as already applied filters not impacting value count.  Definitely something that should be addressed with continued improvement of this component, better event callbacks and filter state management between FilterTable and MetaFilter.

frontend/src/components/filtertable.js
```diff
@@ -160,3 +169,137 @@ export class FilterTable extends React.Component {
...
+  updateValueOptions = () => {
+    const {fieldSelection} = this.state
+    const {customFilters} = this.props
+
+    console.log('CUSTOMFILTER: '+customFilters)
+    if (fieldSelection !== null) {
+      let api_filter = toAPIFilter(customFilters).join()
+      console.log('APIFILTER: '+customFilters)
+
+      fetch(buildUrl(Settings.serverUrl + '/widget/result-aggregator',
```
rsnyman, line 247: Now that the auth stuff has been merged, this needs to change.

```suggestion
      HttpClient.get([Settings.serverUrl, 'widget', 'result-aggregator'], {...})
          .then(response => HttpClient.handleResponse(response))
          .then(data => ....);
```

rsnyman (changes requested):
Now that the Auth PR has been merged, there's an updated way to get data from the API, which I've given an example of in a comment. Also, you're going to need to rebase and fix some conflicts, it seems. Once that's done, we're good to go.

john-dupuy (approved):

rsnyman (approved):

PR #231, opened 2021-11-04
support for additional custom results

RonnyPfannschmidt:
- update tox.ini to current practices
- update to the use of the __wrapped__ api
- add manual and blocked as results

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -23,7 +23,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python-version: [ "3.8", "3.9", "3.10", "3.11.0-alpha - 3.11"]
+        python-version: [ "3.8", "3.9", "3.10", ] # five is broken on 3.11 - celery update needed "3.11.0-alpha - 3.11"]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -23,7 +23,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: [ "3.8", "3.9", "3.10", "3.11.0-alpha - 3.11"]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
--- a/backend/Containerfile
+++ b/backend/Containerfile
@@ -1,0 +1,16 @@
+FROM  registry.access.redhat.com/ubi8/python-38 as base
+USER 0
+RUN mkdir /ibutsu_venv
+RUN chown -R 1001:0 /ibutsu_venv /srv
+USER 1001
+WORKDIR  /srv
+RUN python -m venv /ibutsu_venv && \
+     /ibutsu_venv/bin/pip install -U pip setuptools wheel
+
+
+FROM base as installed
+
+COPY .  /srv/
+RUN /ibutsu_venv/bin/pip install -e .
+ENTRYPOINT [ "/bin/bash", "/srv/docker/entrypoint.sh" ]
+CMD ["help"]
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -59,7 +59,7 @@
     "yarn": "^1.22.0"
   },
   "scripts": {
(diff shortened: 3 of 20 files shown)


ibutsu_pod.yaml
```diff
@@ -0,0 +1,121 @@
...
+      requests:
+        memory: "128Mi"
+        cpu: "250m"
+      limits:
+        memory: "512Mi"
+        cpu: "500m"
+    volumeMounts:
+    - name: ibutsu-database
+      mountPath: /var/lib/postgresql/data
+    env: &all-env
```
rsnyman, line 23: I get that this is easier, but you're essentially exposing secrets to a container that doesn't need them. Hopefully no one would use this in the real world, but it does present an easy way for an attacker to get to other systems.

rsnyman (changes requested):
In general, this is OK, but in future I would prefer separate PRs for each area that you're changing. For now, if you can just address the two other comments I left, then we can merge this.

For instance, changing the pod stuff would be in one, changing some of the URL handling would be in another, and then the actual custom results would be in its own PR.

rsnyman:
Oh, you'll also need to sort out the linting issue.

ibutsu_pod.yaml
RonnyPfannschmidt: yeah, this is only valid in a docker-compose style setup

unfortunately there is no way to clone them more sanely

frontend/package.json
```diff
@@ -59,11 +59,11 @@
     "yarn": "^1.22.0"
   },
   "scripts": {
-    "start": "serve -s build -l tcp://0.0.0.0:8080",
+    "start": "serve -s build -l tcp://0.0.0.0:3000",
```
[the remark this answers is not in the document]
RonnyPfannschmidt: @rsnyman according to the pod shell script, this is the port to use , so we may need to fix that one then

this was a source of confusion, we may need to sort this out differently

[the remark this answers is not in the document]
RonnyPfannschmidt: fixed, was no longer necessary

ibutsu_pod.yaml
RonnyPfannschmidt: split into config maps

rsnyman:
It looks like the [test are failing](https://github.com/ibutsu/ibutsu-server/runs/6673227009?check_suite_focus=true) on the new `:=` operator with Python 3.7. We can probably drop 3.7, but our images are currently based on 3.8 so we'd need to keep that. If we can add in 3.10 testing, that would be cool.

.github/workflows/tests.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [ 3.8, 3.9, '3.10', '3.11']
```
rsnyman, line 26: It looks like GitHub doesn't support Python 3.11 yet.

RonnyPfannschmidt: fixed

backend/Containerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR  /srv
+RUN python -m venv /ibutsu_venv && \
+     /ibutsu_venv/bin/pip install -U pip setuptools wheel
+
+
+FROM base as installed
+
+COPY .  /srv/
+RUN /ibutsu_venv/bin/pip install -e .
+ENTRYPOINT [ "/bin/bash", "/srv/docker/entrypoint.sh" ]
```
quarckster, line 15: I'm wondering why don't you use `dumb-init` here?

RonnyPfannschmidt: the entrypoint here is a shell that runs actions and then starts subprocesses, as such it correctly handles the signals

if i where to exec into something, then dumb init would be needed

rsnyman:
```
         # If you have environment variables, like when running on OpenShift, create the db url
        maybe_db_uri = maybe_sql_url(config.get_namespace("POSTGRESQL_")) or maybe_sql_url(
            config.get_namespace("POSTGRES_")
        )
>       assert maybe_db_uri is not None
E       assert None is not None
```

backend/Containerfile
RonnyPfannschmidt: good catch, will prepare a followup

rsnyman (approved):

Issue #235, opened 2021-11-11
[auth] get_run_list and get_result_list don't check if a user is a part of the project


john-dupuy:
Steps to reproduce:
1) Have 1 projects in your DB
2) Have 1 or more runs/results in your DB that belong to the project
3) Add a user that is not a part of the project 
4) Navigate to runs/results page

Expected results:
4) No runs/results should be displayed, since the user is not a part of the project 

Actual results:
4) Runs/Results are displayed, but clicking on them gives a 403 because get_run and get_result do check if a user is a part of the project. 

Note: this is a follow-up from #227

PR #237, opened 2021-11-11
Modify add_user_filter to filter on User projects

john-dupuy:
Addresses #235

- Ensure a project owner is added to the project when it's created

### DIFF
--- a/backend/ibutsu_server/controllers/result_controller.py
+++ b/backend/ibutsu_server/controllers/result_controller.py
@@ -98,16 +98,15 @@ def get_result_list(filter_=None, page=1, page_size=25, estimate=False, token_in
     user = User.query.get(user)
     query = Result.query
     if user:
-        query = add_user_filter(query, user)
+        query = add_user_filter(query, user, model=Result)
+
     if filter_:
         for filter_string in filter_:
             filter_clause = convert_filter(filter_string, Result)
             if filter_clause is not None:
                 query = query.filter(filter_clause)
 
-    if estimate and not filter_:
-        total_items = get_count_estimate(query, no_filter=True, tablename="results")
-    elif estimate:
+    if estimate:
         total_items = get_count_estimate(query)
     else:
         total_items = query.count()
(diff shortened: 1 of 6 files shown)


backend/ibutsu_server/controllers/result_controller.py
```diff
@@ -98,16 +98,15 @@ def get_result_list(filter_=None, page=1, page_size=25, estimate=False, token_in
...
-        query = add_user_filter(query, user)
+        query = add_user_filter(query, user, filter_=filter_, model=Result)
+
     if filter_:
         for filter_string in filter_:
             filter_clause = convert_filter(filter_string, Result)
             if filter_clause is not None:
                 query = query.filter(filter_clause)
 
-    if estimate and not filter_:
```
john-dupuy, line 108: This is partially addressing #238, but there's still some work to be done as the estimated count is still inaccurate.

backend/ibutsu_server/util/projects.py
```diff
@@ -40,4 +41,15 @@ def add_user_filter(query, user, project=None):
         query = query.filter(or_(User.projects.any(id=project.id), project.owner == user))
     else:
         query = query.filter(or_(Project.users.any(id=user.id), Project.owner == user))
+    if model and model.__tablename__ in ["results", "runs"]:
+        if filter_ is None or not any("project_id" in s for s in filter_):
+            if user.projects:
```
john-dupuy, line 46: Should we also check if the user is a project owner?

backend/ibutsu_server/util/projects.py
```diff
@@ -40,4 +41,15 @@ def add_user_filter(query, user, project=None):
...
+                filter_clause = convert_filter(
+                    f"project_id*{';'.join([project.id for project in user.projects])}", model
+                )
+            else:
+                # if no user.projects filter on entities without any project
+                filter_clause = convert_filter("project_id@n", model)
```
john-dupuy, line 52: Maybe better to force filtering a project?

rsnyman (approved):

Issue #236, opened 2021-11-11
[auth] API should provide a method for owners to add users to a project


john-dupuy:
Currently, users must be added to the project by a DB query. We should have an API endpoint that allows project owner's to do this.

PR #240, opened 2021-11-17
Support adding users to projects via the API

john-dupuy:
Allow users to be added to projects by the API:
- This is done by sending a list of usernames (emails) in the request body, e.g.

```json
{
  "users": ["jdupuy@redhat.com", "jdoe@blah.com", "jdeer@redhat.com"]
}
```
If the user does not exist in the DB, they won't be added. If the user is already in the project, they won't be added. 

Users can only update projects under the following conditions:
- They are a project owner
- They are a super admin

Depends on #237 
Addresses #236

### DIFF
--- a/backend/ibutsu_server/controllers/project_controller.py
+++ b/backend/ibutsu_server/controllers/project_controller.py
@@ -99,11 +99,13 @@ def update_project(id_, project=None, token_info=None, user=None):
     if not is_uuid(id_):
         id_ = convert_objectid_to_uuid(id_)
     project = Project.query.get(id_)
-    user = User.query.get(user)
-    if project and project.owner.id != user.id:
-        return "Forbidden", 403
+
     if not project:
         return "Project not found", 404
+
+    user = User.query.get(user)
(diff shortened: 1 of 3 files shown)


backend/ibutsu_server/controllers/project_controller.py
```diff
@@ -99,12 +99,24 @@ def update_project(id_, project=None, token_info=None, user=None):
...
-    user = User.query.get(user)
-    if project and project.owner.id != user.id:
-        return "Forbidden", 403
+
     if not project:
         return "Project not found", 404
-    project.update(connexion.request.get_json())
+
+    user = User.query.get(user)
+    if not user.is_superadmin:
```
rsnyman, line 107: You could collapse this into a single `if`

```suggestion
    if not user.is_superadmin and project and project.owner.id != user.id:
```

rsnyman (changes requested):
Just one minor thing.

john-dupuy:
The frontend build appears to be failing due to https://github.com/greatSumini/react-facebook-login/issues/74

rsnyman (approved):

PR #245, opened 2021-11-19
Some code refactoring after static analysis.

rsnyman:
- Remove old script that is no longer needed
- Reduce cognitive complexity in auth method
- Reduce cognitive complexity in auth method
- Reduce cognitive complexity in convert_filter method
- Collapse an if in tasks/__init__.py
- Reduce cognitive complexity in run_junit_import method
- Reduce cognitive complexity in run_archive_import method
- Fix more issues from static code analysis
- Remove some code that would never be executed
- Fix up a couple of JS issues


backend/ibutsu_server/scripts/mongo2postgres.py
```diff
@@ -1,425 +0,0 @@
-#!/bin/env python3
```
john-dupuy, line 1: With this gone can we remove the dep on `pymongo`?

rsnyman: Whoops, didn't see your comment till now. Yes we can, I'll do that here too.

rsnyman:
I had to add the `pymongo` dependency back in because we still have some legacy ID checking going on. We can remove it, but I'll leave that to another PR.

PR #248, opened 2021-11-22
Add ability to create superadmin user on startup

john-dupuy:
Create a super admin given environment variables. 

- [x] Update the template to generate a superadmin

I built this image locally and tagged it as `quay.io/ibutsu/backend:add-superadmin-startup` so we can test it on the staging server.

### DIFF
--- a/backend/ibutsu_server/__init__.py
+++ b/backend/ibutsu_server/__init__.py
@@ -13,6 +13,7 @@ from ibutsu_server.db.base import db
 from ibutsu_server.db.base import session
 from ibutsu_server.db.models import upgrade_db
 from ibutsu_server.db.models import User
+from ibutsu_server.db.util import add_superadmin
 from ibutsu_server.encoder import JSONEncoder
 from ibutsu_server.tasks import create_celery_app
 from ibutsu_server.util.jwt import decode_token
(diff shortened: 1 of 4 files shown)


backend/ibutsu_server/__init__.py
```diff
@@ -40,7 +41,7 @@ def get_app(**extra_config):
 
     # Shortcut
     config = app.app.config
-    config.setdefault("BCRYPT_LOG_ROUNDS", 20)
+    config.setdefault("BCRYPT_LOG_ROUNDS", 12)
```
john-dupuy, line 44: Having the log rounds be `20` made it take around 45 - 60 s just to create a user. It also took around 45 - 60 s to check the password of the user.

rsnyman: Yeah, thanks, I was going to change this :-)

backend/ibutsu_server/db/util.py
```diff
@@ -29,3 +33,26 @@ def pg_explain(element, compiler, **kw):
...
+    elif user and not user.is_superadmin:
+        user.is_superadmin = True
+    else:
+        user = User(
+            email=admin_user["email"],
+            name=admin_user.get("name") or "Ibutsu Admin",
+            password=admin_user["password"],
+            is_superadmin=True,
+            is_active=True,
+            activation_code=urlsafe_b64encode(str(uuid4()).encode("utf8")).strip(b"=").decode(),
```
rsnyman, line 54: The activation code ~~can~~ should be blank if the user is active.

rsnyman (changes requested):

rsnyman (approved):

PR #251, opened 2021-11-24
Add a task for adding users/project owners

john-dupuy:
Provide a way for Ibutsu superadmins to seed users and pair them with projects.

This should be merged after #249.

### DIFF
--- a/backend/ibutsu_server/tasks/db.py
+++ b/backend/ibutsu_server/tasks/db.py
@@ -4,8 +4,10 @@ from datetime import timedelta
 
 from ibutsu_server.db.base import session
 from ibutsu_server.db.models import Artifact
+from ibutsu_server.db.models import Project
 from ibutsu_server.db.models import Result
 from ibutsu_server.db.models import Run
+from ibutsu_server.db.models import User
 from ibutsu_server.tasks import task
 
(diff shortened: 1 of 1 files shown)


backend/ibutsu_server/tasks/db.py
```diff
@@ -82,3 +84,79 @@ def prune_old_runs(months=12):
...
+                project_owner = User.query.filter_by(email=project_info["owner"]).first()
+                if not project_owner:
+                    project_owner = User(
+                        email=project_info["owner"],
+                        name=project_info["owner"].split("@")[0],
+                        is_active=True,
+                    )
+                session.add(project_owner)
+                project.owner = project_owner
+                session.add(project)
```
rsnyman, line 143: As far as I know, you actually only need one of these `add`s. If you just `add` the project, then the user will be created and then added to the project automatically.

rsnyman (changes requested):

backend/ibutsu_server/tasks/db.py
john-dupuy: Huh, good to know, let me update that

rsnyman (approved):

Issue #252, opened 2021-11-29
Add JWT_SECRET to ocp config


john-dupuy:
Generate and add `JWT_SECRET` to the OCP template config. The secret should be handled in the same manner as the IBUTSU_SUPERADMIN parameters.

PR #256, opened 2021-11-30
Adjustments for app-sre deployment

bsquizz:


backend/ibutsu_server/__init__.py
```diff
@@ -35,6 +35,13 @@ def _make_sql_url(hostname, database, **kwargs):
     return "postgresql://{}/{}".format(url, database)
 
 
+def _make_broker_url(env_var_value, hostname, password, port):
+    if env_var_value:
+        return env_var_value
+    user_pass_str = ":{}@".format(password) if password else ""
+    return "redis://{}{}:{}".format(user_pass_str, hostname, port)
```
john-dupuy, line 42: Can we change this to use `f`-strings?

ocp-templates/app-interface/flower.yaml
```diff
@@ -0,0 +1,69 @@
...
+          deploymentConfig: celery-flower
+      spec:
+        containers:
+        - env:
+          # TOOD: use redis secret to set this
+          - name: BROKER_URL
+            value: redis://:${REDIS_PASSWORD}@redis.${NAMESPACE}.svc
+          image: celery-flower
+          imagePullPolicy: IfNotPresent
+          name: quay.io/ibutsu/celery-flower:${IMAGE_TAG}
```
john-dupuy, line 35:
```suggestion
          image: quay.io/ibutsu/flower:${IMAGE_TAG}
          imagePullPolicy: Always
```

bsquizz: Adjusted image name but the pull policy should stay at 'IfNotPresent' -- it's a recommended best practice for AppSRE

ocp-templates/app-interface/backend.yaml
```diff
@@ -0,0 +1,161 @@
...
+    strategy:
+      type: Rolling
+    template:
+      metadata:
+        labels:
+          app: ${APP_NAME}
+          deploymentConfig: ibutsu-backend
+      spec:
+        containers:
+        - env:
```
john-dupuy, line 29: I'm happy to do this as a follow-up, but there is one more secret env var we need to run Ibutsu + Auth:
https://github.com/ibutsu/ibutsu-server/issues/252

backend/ibutsu_server/__init__.py
```diff
@@ -64,7 +71,19 @@ def get_app(**extra_config):
...
+                    config.get("REDIS_HOSTNAME"),
+                    config.get("REDIS_PASSWORD"),
+                    config.get("REDIS_PORT"),
+                ),
+                "CELERY_RESULT_BACKEND": _make_broker_url(
+                    config.get("CELERY_RESULT_BACKEND"),
+                    config.get("REDIS_HOSTNAME"),
+                    config.get("REDIS_PASSWORD"),
+                    config.get("REDIS_PORT"),
+                ),
```
john-dupuy, line 86: Can we move these into their own `if` block? 

```python
if config.get("CELERY_BROKER_URL") or config.get("REDIS_HOSTNAME"):
```

```python
if config.get("CELERY_RESULT_BACKEND") or config.get("REDIS_HOSTNAME"):
```

bsquizz: Ah, didn't catch that. Yep.

john-dupuy (approved):

bsquizz:
@john-dupuy @rsnyman you guys ok w/ merging this? I think the main impact would be just making sure that the value of the CELERY_BROKER_URL still gets read in and used for redis as expected on the existing PSI deployment

john-dupuy:
Yup, I think we're good to merge this in.

PR #266, opened 2021-12-09
Show login progress feedback

rsnyman:
- Add a spinner to the login button so that people can see that something is happening


john-dupuy:
Nice - that's slick!

PR #268, opened 2021-12-10
Add JWT secret to template files

john-dupuy:
Address #252

### DIFF
--- a/ocp-templates/app-interface/backend.yaml
+++ b/ocp-templates/app-interface/backend.yaml
@@ -91,6 +91,11 @@ objects:
               secretKeyRef:
                 key: name
                 name: ibutsu-superadmin
+          - name: JWT_SECRET
+            valueFrom:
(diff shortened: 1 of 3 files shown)


ocp-templates/app-interface/backend.yaml
```diff
@@ -91,6 +91,11 @@ objects:
               secretKeyRef:
                 key: name
                 name: ibutsu-superadmin
+          - name: JWT_SECRET
```
john-dupuy, line 94: @bsquizz what do we need to do on the AppSRE side? Add it to vault?

bsquizz: Yep, we will add it to vault and then update the app-interface config for our namespace to insert the secret into the ns.

rsnyman (approved):

rsnyman (approved):

PR #270, opened 2021-12-15
Fix keycloak login

rsnyman:
The arguments were in the wrong place. Not sure why it worked locally, but now it is working against the stage SSO environment here.


john-dupuy (approved):
LGTM - tested locally with stage auth and it's working.

PR #276, opened 2022-01-12
Add test history tab to Result page

john-dupuy:
Adding a test result history tab to the `ResultView`. This page contains the following:
- Dropdown to select a date range (1 Week, 2 Weeks, 1 Month, 2 Months, 3 Months, 5 Months) from which to view the test's history
- Quick filter to view only failed/error results
- Small summary of passed/failed/error given the date range (note this ignores the quick filter for only viewing failed/error results)
- Table of historical results

It looks like this:
Screenshot from 2022-01-13 14-23-36

It is also possible to view this page within the classify-failures page of a run:
Screenshot from 2022-01-13 14-25-18

I also introduce lazy-loading of the `ResultView` on table expansion for both this page and the `classify-failures` page. It's largely based on: https://github.com/patternfly/patternfly-react/issues/1450

Especially for failed tests, one potential enhancement would be to show the date the test last passed. 

This work is for: https://issues.redhat.com/browse/IBUTSU-64

### DIFF
--- a/backend/ibutsu_server/__init__.py
+++ b/backend/ibutsu_server/__init__.py
@@ -63,7 +63,7 @@ def get_app(**extra_config):
     config.from_mapping(os.environ)
     # convert str to bool for USER_LOGIN_ENABLED
     if isinstance(config.get("USER_LOGIN_ENABLED", True), str):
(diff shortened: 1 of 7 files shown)


rsnyman (approved):
LGTM

PR #285, opened 2022-01-20
Upgrade celery to handle a security issue, let's see if this works

rsnyman:
We previously pinned Celery due to a bug in Kombu. Many moons have gone by since the advent of this bug and subsequent pin. Let us see if the bug still exists.

### DIFF
--- a/backend/celery_worker.sh
+++ b/backend/celery_worker.sh
@@ -1,1 +1,1 @@
-celery worker -E -A ibutsu_server.tasks.queues:app --no-color
+celery --app ibutsu_server.tasks.queues:app --no-color worker --events
--- a/backend/docker/Dockerfile.scheduler
+++ b/backend/docker/Dockerfile.scheduler
@@ -13,4 +13,4 @@ RUN chgrp -R 0 ibutsu_server && chmod -R g+rwX ibutsu_server
 ENTRYPOINT ["celery"]
 
 # Run application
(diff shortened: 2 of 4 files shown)


john-dupuy (changes requested):
Trying to start the celery worker locally with a command like:
```
celery worker -E -A ibutsu_server.tasks.queues:app --concurrency 1 --loglevel info
```
I get an error:
```
You are using `-A` as an option of the worker sub-command:
celery worker -A celeryapp <...>

The support for this usage was removed in Celery 5.0. Instead you should use `-A` as a global option:
celery -A celeryapp worker <...>
```

So it sounds like we'll need to update our `Dockerfile.worker`?

john-dupuy:
Changing the `-A <app>` part to be after `celery` fixed up worker/scheduler, both appear to be working fine locally after that.

john-dupuy:
I think these two lines will also need updating - https://github.com/ibutsu/ibutsu-server/blob/master/backend/docker/Dockerfile.scheduler#L16 and https://github.com/ibutsu/ibutsu-server/blob/master/backend/docker/Dockerfile.worker#L15

john-dupuy (approved):

rsnyman:
Thanks @john-dupuy you're a legend! Hope you're doing well!

john-dupuy:
No problem! Hope you and the team are doing great too! :smiley: :

PR #288, opened 2022-01-26
Adding project setup fixes to documentation

Griffin-Sullivan:
I was unable to start the project without the settings.yaml symlink in `backend/`. I'm adding that step to setup and also adding .com to the login credentials for clarity.


Griffin-Sullivan:
Added the `export` statement, as I was unable to start up the front end without it.

PR #294, opened 2022-03-01
Fix metadata reading during test import

obaranov:
Looks like the run dict exposes the "metadata" property instead of data. Fixing that to properly import tests results. 

Fixes: https://github.com/ibutsu/ibutsu-server/issues/291


rsnyman:
Ibutsu uses `pre-commit` for enforcing `black` and other linting rules. Please run `pre-commit run --all` against your code to fix the failing lint action. Thank you!

PR #313, opened 2022-03-29
Add basic user management

rsnyman:
- Add a user management endpoint
- Add a project management endpoint
- Add an administration section to the frontend
- Add a user management section to admin frontend
- Refactor headers into a single component
- Refactor page layout into a single component
- Update pre-commit
- Other minor updates and bug fixes

PR #314, opened 2022-03-30
Add data persistence to ibutsu-pod.sh script

rsnyman:
- Add ability to specify options in ibutsu-pod.sh
- Detect if script is running in the right directory
- Add options to the script for creating admin and project
- Add option to enable data persistence in postgres and redis

### DIFF
--- a/scripts/ibutsu-pod.sh
+++ b/scripts/ibutsu-pod.sh
@@ -1,8 +1,115 @@
 #!/bin/bash
+
+# Some variables
 POD_NAME="ibutsu"
 JWT_SECRET=$(cat /dev/urandom | env LC_CTYPE=C tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+IS_PERSISTENT=false
+USE_VOLUMES=false
+CREATE_ADMIN=false
+CREATE_PROJECT=false
+POSTGRES_EXTRA_ARGS=
+REDIS_EXTRA_ARGS=
+BACKEND_EXTRA_ARGS=
 
+function print_usage() {
+    echo "Usage: ibutsu-pod.sh [-h|--help] [-p|--persistent] [-V|--use-volumes] [-A|--create-admin] [-P|--create-project] [POD_NAME]"
+    echo ""
+    echo "optional arguments:"
+    echo "  -h, --help            show this help message"
+    echo "  -p, --persistent      persist the data in the containers"
+    echo "  -V, --use-volumes     use podman volumes to store data"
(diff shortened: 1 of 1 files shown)


rsnyman:
@Griffin-Sullivan take a look at this

scripts/ibutsu-pod.sh
```diff
@@ -12,13 +82,15 @@ podman run -dt \
        -e POSTGRES_USER=ibutsu \
        -e POSTGRES_DB=ibutsu \
        -e POSTGRES_PASSWORD=ibutsu \
+       $POSTGRES_EXTRA_ARGS \
```
Griffin-Sullivan, line 85: At https://hub.docker.com/_/postgres they also recommend storing the data under another folder in `/var/lib/postgres/data`. See the section titled `PGDATA`. We could easily add that with this var.

scripts/ibutsu-pod.sh
```diff
@@ -1,7 +1,77 @@
...
+# Persist the data
+if [[ $IS_PERSISTENT = true ]]; then
+    POSTGRES_EXTRA_ARGS+=' -v ./.postgres-data:/var/lib/postgres/data:Z'
+    REDIS_EXTRA_ARGS+=' -v ./.redis-data:/data:Z'
+    if [[ ! -d ".postgres-data" ]]; then
+        mkdir .postgres-data
+    fi
+    if [[ ! -d ".redis-data" ]]; then
+        mkdir .redis-data
+    fi
```
Griffin-Sullivan, line 66: Could also make these volumes with the `podman volume create` command and use them in vars above.
```bash
if volumes don't exist:
    podman volume create postgres-data
    podman volume create redis-data
POSTGRES_EXTRA_ARGS+=' -v postgres-data:/var/lib/postgres/data:Z'
REDIS_EXTRA_ARGS+=' -v redis-data:/data:Z'
```
Sorry I've never put in the time to understand bash if statements....

rsnyman: I've never been a huge fan of volumes. They're a bit opaque, and can sometimes cause issues when you don't realise that there's a volume with the same name somewhere else.

scripts/ibutsu-pod.sh
rsnyman: OK, I've specified the PGDATA variable.

scripts/ibutsu-pod.sh
rsnyman: I added an option to use volumes.

rsnyman:
The checks are failing due to external issues, and these issues are already solved in #313, so I've marked it as a draft.

PR #328, opened 2022-04-14
Check if the first argument passed to on_failure is a dict

lina-is-here:
I'm continuing chasing down errors in logs. I tried a few things but couldn't reproduce the error below. Therefore adding more logging to understand what's going on and what kind of data comes here. 
```
[2022-04-13 23:55:13,287: ERROR/MainProcess] Task handler raised error: AttributeError("'str' object has no attribute 'get'")
Traceback (most recent call last):
  File "/opt/app-root/lib64/python3.8/site-packages/billiard/pool.py", line 362, in workloop
    result = (True, prepare_result(fun(*args, **kwargs)))
  File "/opt/app-root/lib64/python3.8/site-packages/celery/app/trace.py", line 555, in _fast_trace_task
    R, I, T, Rstr = tasks[task].__trace__(
  File "/opt/app-root/lib64/python3.8/site-packages/celery/app/trace.py", line 499, in trace_task
    I, _, _, _ = on_error(task_request, exc, uuid)
  File "/opt/app-root/lib64/python3.8/site-packages/celery/app/trace.py", line 338, in on_error
    R = I.handle_error_state(
  File "/opt/app-root/lib64/python3.8/site-packages/celery/app/trace.py", line 167, in handle_error_state
    return {
  File "/opt/app-root/lib64/python3.8/site-packages/celery/app/trace.py", line 220, in handle_failure
    task.on_failure(exc, req.id, req.args, req.kwargs, einfo)
  File "/app/ibutsu_server/tasks/__init__.py", line 53, in on_failure
    report_id = args[0].get("id")  # get the report ID from the task
AttributeError: 'str' object has no attribute 'get'
```

### DIFF
--- a/backend/ibutsu_server/tasks/__init__.py
+++ b/backend/ibutsu_server/tasks/__init__.py
@@ -48,15 +48,21 @@ def create_celery_app(_app=None):
 
         def on_failure(self, exc, task_id, args, kwargs, einfo):
             # if the task is related to a report, set that report to failed
-            try:
-                with _app.app_context():
-                    report_id = args[0].get("id")  # get the report ID from the task
-                    report = Report.query.get(report_id)
-                    # if this is actually a report ID, set it as an error
-                    if report:
-                        self._set_report_error(report)
-            except (IndexError, TypeError):
-                pass
+            if isinstance(args[0], dict):
+                try:
+                    with _app.app_context():
+                        report_id = args[0].get("id")  # get the report ID from the task
+                        report = Report.query.get(report_id)
+                        # if this is actually a report ID, set it as an error
+                        if report:
+                            self._set_report_error(report)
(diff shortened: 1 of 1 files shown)


rsnyman:
I think I know what is causing this. Let me see if I can track it down.

rsnyman:
So what's happening here is that this method was created when the only thing that Celery was used for was the reports. We were experiencing task failures, and were trying to mark the report as error'd so that the user has a nicer experience. To do this, we tried to catch the failure and then get the report id out of the arguments.

Now we have other tasks, the first argument is no longer a dictionary, but could be a str or int, or something else.

So the solution is probably to just check if the first argument is a dictionary.

lina-is-here:
Thanks @rsnyman for the explanation! I will modify this PR

PR #348, opened 2022-05-19
Add Compare Runs view

Griffin-Sullivan:
This feature adds the ability for admins to define a new view for comparing runs. At the moment users can use MetaFilter and FilterTable to select the different types of results they would like to compare. They will then retrieve a list of tests where the results are not the same. I have a video demo I can share if you would like to see current functionality while this is being reviewed.

Some relevant notes:
- I made an update to MetaFilter so that the filters are shown within the MetaFilter render function instead of the FilterTable's render
- I also gave MetaFilter a property called `id` that takes a number so that you can define multiple filters for your FilterTable
- Adjusted ResultView to allow optional showing of artifacts (logs)
- Adjusted ResultView and TestHistory to take a property `comparisonResults`, so that I can pass along only the results I retrieved from the DB to the `onCollapse` function in `compareruns.js`

### DIFF
--- a/backend/ibutsu_server/controllers/widget_controller.py
+++ b/backend/ibutsu_server/controllers/widget_controller.py
@@ -1,6 +1,7 @@
 import connexion
 from ibutsu_server.constants import ALLOWED_TRUE_BOOLEANS
 from ibutsu_server.constants import WIDGET_TYPES
+from ibutsu_server.widgets.compare_runs_view import get_comparison_data
 from ibutsu_server.widgets.jenkins_heatmap import get_jenkins_heatmap
 from ibutsu_server.widgets.jenkins_job_analysis import get_jenkins_analysis_data
 from ibutsu_server.widgets.jenkins_job_analysis import get_jenkins_bar_chart
@@ -11,6 +12,7 @@ from ibutsu_server.widgets.result_summary import get_result_summary
 from ibutsu_server.widgets.run_aggregator import get_recent_run_data
 
 WIDGET_METHODS = {
+    "compare-runs-view": get_comparison_data,
     "jenkins-analysis-view": get_jenkins_analysis_data,
     "jenkins-bar-chart": get_jenkins_bar_chart,
     "jenkins-heatmap": get_jenkins_heatmap,
@@ -49,7 +51,11 @@ def _typecast_params(widget_id, params):
             params[param] = params[param].lower()[0] in ALLOWED_TRUE_BOOLEANS
         elif param in param_types and param_types[param] == "float":
             params[param] = float(params[param])
-        elif param in param_types and param_types[param] == "list":
+        elif (
+            param in param_types
+            and param_types[param] == "list"
+            and not isinstance(params[param], list)
+        ):
             params[param] = params[param].split(",")
     return params
 
@@ -79,6 +85,9 @@ def get_widget(id_):
     """
     if id_ not in WIDGET_TYPES.keys():
         return "Widget not found", 404
-    params = _pre_process_params(dict(connexion.request.args))
+    params = {}
+    for key in connexion.request.args.keys():
+        params[key] = connexion.request.args.getlist(key)
+    params = _pre_process_params(params)
     params = _typecast_params(id_, params)
     return WIDGET_METHODS[id_](**params)
--- a/backend/ibutsu_server/filters.py
+++ b/backend/ibutsu_server/filters.py
@@ -29,6 +29,7 @@ OPER_COMPARE = {
 }
 FILTER_RE = re.compile(r"([a-zA-Z\._]+)([" + "".join(OPERATORS.keys()) + "])(.*)")
 FLOAT_RE = re.compile(r"[+-]?(\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?")
+VERSION_RE = re.compile("([0-9].*[0-9])")
 
 
 def _to_int_or_float(value):
@@ -100,6 +101,7 @@ def convert_filter(filter_string, model):
     field = match.group(1)
     oper = match.group(2)
     value = match.group(3).strip('"')
+    is_version = True if VERSION_RE.match(value) is not None else False
     column = string_to_column(field, model)
     # determine if the field is an array field, if so it requires some additional care
     is_array_field = field in ARRAY_FIELDS
@@ -108,10 +110,9 @@ def convert_filter(filter_string, model):
         return _null_compare(column, value)
     elif oper == "*":
         value = value.split(";")
-    elif "build_number" not in field:
+    elif not is_version and "build_number" not in field:
         # This is a horrible hack, because Jenkins build numbers are strings :-(
         value = _to_int_or_float(value)
-
     if is_array_field:
         return _array_compare(oper, column, value)
     else:
--- a/backend/ibutsu_server/test/test_widget_controller.py
+++ b/backend/ibutsu_server/test/test_widget_controller.py
@@ -1,0 +1,87 @@
+from datetime import datetime
+from unittest.mock import patch
+
+from ibutsu_server.test import BaseTestCase
+from ibutsu_server.test import MockProject
+from ibutsu_server.test import MockResult
+
+MOCK_ID_1 = "99fba7d2-4d32-4b9b-b07f-4200c9717661"
+MOCK_ID_2 = "99fba7d2-4d32-4b9b-b07f-4200c9717662"
+MOCK_RUN_ID_1 = "8395c072-4724-4dae-9f4a-1407b25a2d5b"
+MOCK_RUN_ID_2 = "8395c072-4724-4dae-9f4a-1407b25a2d5c"
+MOCK_PROJECT = MockProject()
+START_TIME = datetime.utcnow()
+MOCK_RESULTS = [
+    MockResult(
+        id=MOCK_ID_1,
+        duration=6.027456183070403,
+        result="passed",
+        component="fake-component",
+        data={
+            "jenkins_build": 145,
+            "commit_hash": "F4BA3E12",
+            "assignee": "jdoe",
+            "component": "fake-component",
+            "fspath": "fake_component.py",
+        },
+        start_time=str(START_TIME),
+        source="source",
+        params={"provider": "vmware", "ip_stack": "ipv4"},
+        test_id="test_id",
+        project=MOCK_PROJECT,
+    ),
+    MockResult(
+        id=MOCK_ID_2,
+        duration=10.457123453030907,
+        result="failed",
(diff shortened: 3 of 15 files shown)


backend/ibutsu_server/widgets/compare_runs_view.py
```diff
@@ -0,0 +1,189 @@
...
+                },
+                "total_execution_time": datum.total_execution_time,
+            }
+        )
+
+    return data
+
+
+def get_compare_runs_view(filter_=None, project=None, page=1, page_size=25, run_limit=None):
+    print('hello')
```
mshriver, line 119: hey there!

Griffin-Sullivan: You know how the initial debugging goes lol

backend/ibutsu_server/filters.py
```diff
@@ -99,7 +102,17 @@ def convert_filter(filter_string, model):
...
+    is_string = False
+    is_version = False
+    match = VERSION_RE.match(value)
+    if match:
+        value = match.group(1)
+        is_version = True
+    match = STRING_RE.match(value)
+    if match:
+        value = match.group(1)
+        is_string = True
```
Griffin-Sullivan, line 115: Changes here were part of a debugging session with @rsnyman where we are trying to fix a problem with comparing BaseOS. Still need to revisit.

backend/ibutsu_server/widgets/compare_runs_view.py
```diff
@@ -0,0 +1,189 @@
...
+from ibutsu_server.db.models import Result
+from ibutsu_server.filters import apply_filters
+from ibutsu_server.filters import convert_filter
+from ibutsu_server.filters import string_to_column
+from ibutsu_server.util.count import get_count_estimate
+from ibutsu_server.util.projects import add_user_filter
+from sqlalchemy import desc
+from sqlalchemy import func
+
+def _get_jenkins_aggregation(filters=None, project=None, page=1, page_size=25, run_limit=None):
```
Griffin-Sullivan, line 21: This was copy and paste from how the Jenkins Job widget-config works. @rsnyman is this necessary? I can't even tell if this ever gets run. If so it needs to be more personalized for comparison widget cause this is kind of an unknown hack to me.

frontend/src/views/compareruns.js
```diff
@@ -0,0 +1,335 @@
...
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      columns: [{title: 'Test', cellFormatters: [expandable]}, 'Run 1', 'Run 2'],
+      rows: [getSpinnerRow(3)],
+      results: [],
+      selectedResults: [],
+      cursor: null,
+      pageSize: 1000,
```
Griffin-Sullivan, line 51: Is there anyway to set this as unlimited?

rsnyman (changes requested):

backend/ibutsu_server/widgets/compare_runs_view.py
Griffin-Sullivan: Yeah will do. Jenkins Jobs was the only other full page view I knew about when I started so I thought I needed that to get the widget config up.

backend/ibutsu_server/controllers/widget_controller.py
```diff
@@ -79,6 +83,14 @@ def get_widget(id_):
     """
     if id_ not in WIDGET_TYPES.keys():
         return "Widget not found", 404
-    params = _pre_process_params(dict(connexion.request.args))
-    params = _typecast_params(id_, params)
+    # Restructure data into arrays for comparison-data
+    if id_ == "comparison-data":
```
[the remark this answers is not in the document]
Griffin-Sullivan: I made some changes so it's not hard coded. Let me know what you think.

Griffin-Sullivan:
I got the view in a state that I think is good for review. I'm going to write a test for the backend and then I will take this out of draft!

Griffin-Sullivan:
Lint is failing because line is too long. When I separate it into different lines with parenthesis, python black changes it all to one line, which is too long for linter.

backend/ibutsu_server/test/test_widget_controller.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @patch("ibutsu_server.widgets.compare_runs_view.Result.query")
+    def test_get_comparison_result_list(self, mocked_query):
+        """Test case for compare_runs_view.py::get_comparison_data
+
+        Get the list of comparison results.
+        """
+        MOCK_RUN_IDS = [MOCK_RUN_ID_1, MOCK_RUN_ID_2]
+        MOCKED_RESULTS = [[mocked_result] for mocked_result in MOCK_RESULTS]
+        mocked_query = mocked_query.filter.return_value
```
Griffin-Sullivan, line 67: These lines were added to avoid creating too long of a line for the linter on the lines below.

backend/ibutsu_server/controllers/widget_controller.py
```diff
@@ -79,6 +85,10 @@ def get_widget(id_):
     """
     if id_ not in WIDGET_TYPES.keys():
         return "Widget not found", 404
-    params = _pre_process_params(dict(connexion.request.args))
-    params = _typecast_params(id_, params)
+    else:
```
rsnyman, line 88: I'm not sure why you added the `else` here? The `if` section will return, so the `else` is unnecessary, hence why it was left out originally.

frontend/src/components/filtertable.js
```diff
@@ -32,17 +32,13 @@ export class FilterTable extends React.Component {
     rows: PropTypes.array,
     actions: PropTypes.array,
     filters: PropTypes.array,
-    activeFilters: PropTypes.object,
-    hideFilters: PropTypes.array,
```
rsnyman, line 36: We use this table in a few places, and I'm a bit worried about removing this stuff without proper testing. Did you check the runs and results pages?

rsnyman: I'm going to check this again tomorrow (Friday) to see how it affects the other tables.

Griffin-Sullivan: Ok I need to refine some things.

Griffin-Sullivan: As you can see below I thought I could put this back into draft, but I guess not lol. Anyways, I removed all of those changes to FilterTable, so there should be no worries. I checked other pages and they looked good. Just had to modify some stuff on my view. You can get the same functionality I was after if you pass those functions to MetaFilter and not worry about it in FilterTable.

rsnyman:
Also, I'm not entirely sure what each of those filters is supposed to be for or supposed to do. Is there a way you can make the description clearer?

rsnyman:
I added a widget to my dashboard, but now nothing is showing up? I don't see any errors in my console, any ideas what might be causing that?

frontend/src/components/view.js
```diff
@@ -10,9 +10,10 @@ import {
 
 import { HttpClient } from '../services/http';
 import { Settings } from '../settings';
-import { JenkinsJobView, JenkinsJobAnalysisView } from '../views';
+import { CompareRunsView, JenkinsJobView, JenkinsJobAnalysisView } from '../views';
 
 const VIEW_MAP = {
+  'compare-runs-view': CompareRunsView,
```
rsnyman, line 16: Everywhere else your view is called `comparison-data` and then suddenly here you call it `compare-runs-view`. Please use the same name everywhere.

rsnyman (changes requested):

frontend/src/components/view.js
Griffin-Sullivan: `comparison-data` is for returning the data used in the Compare Runs view I created. `compare-runs-view` was for returning the view itself. Do these not need to be separate?

rsnyman: Aaah, OK. I misunderstood - I thought there was only 1 thing in this, the comparison view.

rsnyman:
OK, I took a look at this again, and I'm very much still getting issues.

1. On the dashboard, when I add a new widget, there is a "Comparison Data" widget which has no JS code associated with it in the widgets submodule and therefore doesn't show up on the dashboard. If it is not supposed to show up on the dashboard, we probably need to figure out a way to hide it.
2. When trying to create a new Compare Runs View via the API, when I attempt to add a "compare-runs-view" type, I get the following error message: `Bad request, widget type does not exist`
3. There definitely seems to be some sort of mixup between "compare-runs-view" and "comparison-data"

Griffin-Sullivan:
1. I didn't have the intention of making it show up on the dashboard, so hiding it would be nice or maybe I need to not use a widget for this? Or I could make it an available dashboard widget?
2. I will send you what my config looks like and you can try that. Did you have `type: widget` instead of `type: view`?
3. What would that be?

rsnyman:
This looks much better! There's one minor bug I encountered while adding and updating my widget config, and it would be great if you could include the fix here:

```diff
diff --git a/backend/ibutsu_server/controllers/widget_config_controller.py b/backend/ibutsu_server/controllers/widget_config_controller.py
index 8b03162..09ed1f0 100644
--- a/backend/ibutsu_server/controllers/widget_config_controller.py
+++ b/backend/ibutsu_server/controllers/widget_config_controller.py
@@ -33,9 +33,9 @@ def add_widget_config(widget_config=None, token_info=None, user=None):
             return "Forbidden", 403
         data["project_id"] = project.id
     # default to make views navigable
-    if data.get("navigable"):
+    if data.get("navigable") and isinstance(data["navigable"], str):
         data["navigable"] = data["navigable"][0] in ALLOWED_TRUE_BOOLEANS
-    if data.get("type") == "view" and data.get("navigable") is not None:
+    if data.get("type") == "view" and data.get("navigable") is None:
         data["navigable"] = True
     widget_config = WidgetConfig.from_dict(**data)
     session.add(widget_config)
@@ -121,10 +121,10 @@ def update_widget_config(id_, token_info=None, user=None):
     if not widget_config.weight:
         widget_config.weight = 10
     # default to make views navigable
-    if data.get("navigable"):
+    if data.get("navigable") and isinstance(data["navigable"], str):
         data["navigable"] = data["navigable"][0] in ALLOWED_TRUE_BOOLEANS
-    if data.get("type") and data["type"] == "view" and data.get("navigable") is not None:
-        data.navigable = True
+    if data.get("type") and data["type"] == "view" and data.get("navigable") is None:
+        data["navigable"] = True
     widget_config.update(data)
     session.add(widget_config)
     session.commit()
```

rsnyman (approved):

mshriver:
@Griffin-Sullivan @rsnyman thank you so much for the work that went into adding this result view!

PR #375, opened 2022-07-26
Update podman and Docker Compose configurations

### LINKED ISSUE #362: Can't log in after containerized deployment
When using the build script `./scripts/ibutsu-pod.sh --create-admin --create-project`
After all containers are running and browsing to the Frontend url, there's no option to enter the username and password generated.
Instead there's a message: 

>  *Log in via one of the icons below.*

rsnyman:
- Prevent backend and worker from interfering with each other's virtual environments
- Make the backend listen on the correct port in the Docker containers
- Fixes #362

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,8 @@ package-lock.json
 .ibutsu-env
 .ibutsu_env
 .ibutsu
+.backend_env
+.worker_env
 
 # Frontend
 frontend/public/version.json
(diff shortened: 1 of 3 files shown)


rsnyman:
@lina-is-here I think this will fix your issue too.

lina-is-here:
thanks! I tried these changes and everything works as expected. No errors during the start of the container and I was able to login to the UI as well. Awesome!

lina-is-here (approved):

PR #379, opened 2022-08-08
Promote user_properties on archive import

mshriver:
user_properties is promoted into metadata on results from the client, imported results should get the same treatment.

Tested with an archive with user_properties, need to test with one without as well.

This is not addressing junit import, which would have to pull the `properties` elements when trying to parse metadata.


rsnyman (approved):

mshriver:
thanks @rsnyman !

PR #380, opened 2022-08-09
fix api errors found in fuzzy testing, part 2

lina-is-here:
This is a second batch of fixing API errors found in fuzzy testing.

It includes:
1. Modify the test command - exclude `/api/health` because it returns 5xx error codes
2. Validate UUID in more places
3. Check if the object already exists in various places
4. Replace error code from `500` to `503` in database health check. It is more specific.
5. Fix another typo in openapi schema `description: ID of pet to update`
6. Add `validate_activation_code` function to validate the activation code

### DIFF
--- a/backend/docker/start_fuzz_testing.sh
+++ b/backend/docker/start_fuzz_testing.sh
@@ -11,4 +11,5 @@ response=$(curl -s -X POST http://backend:8080/api/login \
 
 token=$(jq -r '.token' <<< "${response}")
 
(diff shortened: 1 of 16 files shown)


rsnyman (approved):
I was looking at the `validate_uuid` decorator the other day, and thinking, "we need to use that in more places" :grin:

lina-is-here:
@rsnyman yess :D but I also think it would be the best to validate it in the schema :)

Issue #378, opened 2022-08-08
TypeError from floatish filtering


mshriver:
![image](https://user-images.githubusercontent.com/21315076/183414958-59228b72-32ee-4d2c-af27-82144d0f6c88.png)

When filtering metadata fields like `metadata.SomethingVersion` for semantic versions like `1.2.3`, a TypeError is produced, and no results fetched.

@rsnyman suggested these values are getting processed as floats, and the filter is trying to typecast. It should be interpreted as a string.

mshriver:
I'm unable to reproduce the error with local ibutsu having results with this metadata field set (`X.Y`, `X.Y.Z`, and `X.Y.Z.ZZ` formats).

Filtering works as expected

mshriver:
I've now observed different behavior on ibutsu, with no changes to the production containers from when this was written.

The mentioned filters now work intermittently, but will still cause type errors. Still unable to actually reproduce locally, even after loading results that didn't contain the metadata field being searched.

mshriver:
Alright, so some sanity has been found after some digging, learning, and a conversation with @rsnyman.

The root of this issue is the gateway timeout in the haproxy route in OCP.  The backend/api query is taking too long, and the route times out and throws a 504.

The HttpClient functions, and fetch, aren't setup to cleanly handle this, so a TypeError occurs in the frontend.

Three actions are planned in order to better deal with this:
1. Modify the OCP template spec for the backend route to increase the timeout
2. Optimize/Improve performance of the result queries
3. Better handling of potential gateway timeouts in the result-list page

mshriver:
This has been resolved through the debouncing of async/await requests, and backend routing timeout changes in deployment.

PR #381, opened 2022-08-09
Add params logging on query failure

mshriver:
This feels a bit dumb, but the filter behavior has been inconsistent,
and I'd like to be able to see what the params were if/when the failure
occurs. Can use this to reproduce within api tests, as I've been unable to do so (at all)
when running ibutsu locally with a smaller data set.

Related to #378


mshriver:
`This feels a bit dumb` I was right!

PR #391, opened 2022-08-30
fix api errors found in fuzzy testing, part 4

lina-is-here:
This is a 4th batch of fixing API errors found in fuzzy testing.

It includes:

1. Fixing offset in all places
2. Add some more uuid validation
3. Fix `from_dict` method, so it doesn’t throw an error in case of empty keys (e.g. empty strings)
4. Add missing email format validation to openapi schema

### DIFF
--- a/backend/ibutsu_server/controllers/admin/project_controller.py
+++ b/backend/ibutsu_server/controllers/admin/project_controller.py
@@ -6,6 +6,7 @@ from ibutsu_server.db.models import Project
 from ibutsu_server.db.models import User
 from ibutsu_server.filters import convert_filter
 from ibutsu_server.util.admin import check_user_is_admin
+from ibutsu_server.util.query import get_offset
 from ibutsu_server.util.uuid import convert_objectid_to_uuid
 from ibutsu_server.util.uuid import is_uuid
 from ibutsu_server.util.uuid import validate_uuid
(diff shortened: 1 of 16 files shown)


backend/ibutsu_server/controllers/admin/project_controller.py
```diff
@@ -87,10 +88,11 @@ def admin_get_project_list(
     if group_id:
         query = query.filter(Project.group_id == group_id)
 
-    offset = (page * page_size) - page_size
-    offset = 0 if offset < 0 else offset
+    offset = get_offset(page, page_size)
     total_items = query.count()
     total_pages = (total_items // page_size) + (1 if total_items % page_size > 0 else 0)
+    if offset > 9223372036854775807:  # max value of bigint
```
rsnyman, line 94: Is there a constant somewhere with this value?

backend/ibutsu_server/controllers/artifact_controller.py
```diff
@@ -129,6 +131,10 @@ def upload_artifact(body, token_info=None, user=None):
     """
     result_id = body.get("result_id") or body.get("resultId")
     run_id = body.get("run_id") or body.get("runId")
+    if not is_uuid(result_id):
```
rsnyman, line 134: I can't remember what `is_uuid` does for None, but since either of those fields is optional, at least one will be None at any time, and this might throw an exception.

backend/ibutsu_server/controllers/admin/project_controller.py
lina-is-here: oh yeah, good point! I found it in `sys.maxsize`

backend/ibutsu_server/controllers/artifact_controller.py
lina-is-here: looks like it raises `TypeError: one of the hex, bytes, bytes_le, fields, or int arguments must be given`.
I fixed this check, so it happens only if the value is given.

rsnyman (approved):

Issue #390, opened 2022-08-17
Bug: Jenkins bar/line chart not working


jawatts:
Cannot add a Jenkins bar or line chart. In the Satellite QE project, when a user attempts to add a widget, nothing happens. There are no errors that are displayed.

PR #405, opened 2022-11-08
Fixing Jenkins Bar and Line charts

Griffin-Sullivan:
Fixes https://github.com/ibutsu/ibutsu-server/issues/390. Users can now add the Jenkins Bar and Line charts to their dashboards.


frontend/src/widgets/genericarea.js
```diff
@@ -108,6 +108,7 @@ export class GenericAreaWidget extends React.Component {
           areaCharts.push(
             <ChartArea
               data={chartData}
+              key={areaCharts.length}
```
Griffin-Sullivan, line 111: I was getting errors in the UI that there were not unique keys for the `areaCharts` list items. I used the length as something I know would be updated dynamically, but we should use something else here. I was just trying to hack my way to a solution first. Thoughts?

rsnyman: You could just use an increasing index?

Griffin-Sullivan: Updated.

Issue #389, opened 2022-08-17
[Enhance Dashboards] Allow for editing of created widgets


jawatts:
Once a widget is created, there is no way to edit the widget. Allow the user to edit created widgets.

PR #412, opened 2022-11-30
Adding ability to edit dashboard widgets

Griffin-Sullivan:
Solving https://github.com/ibutsu/ibutsu-server/issues/389

This PR adds a button to each widget for editing the widget. When the button is clicked a Modal pops up with the fields already filled out with the widgets details. Users can then make changes and click save. Note: All edits will shown on save except for the widget's title. For some reason it only updates on a full page refresh. I can look into this more. 

Everything else is working correctly, but I have this error in the console that I can't track down in the code. 
Screenshot from 2022-11-30 12-23-05

### DIFF
--- a/frontend/src/dashboard.js
+++ b/frontend/src/dashboard.js
@@ -30,7 +30,7 @@ import {
 import { HttpClient } from './services/http';
 import { KNOWN_WIDGETS } from './constants';
 import { Settings } from './settings';
-import { DeleteModal, NewDashboardModal, NewWidgetWizard } from './components';
+import { DeleteModal, NewDashboardModal, NewWidgetWizard, EditWidgetModal } from './components';
 import {
   GenericAreaWidget,
   GenericBarWidget,
@@ -77,6 +77,8 @@ export class Dashboard extends React.Component {
       isDashboardSelectorOpen: false,
       isNewDashboardOpen: false,
       isWidgetWizardOpen: false,
+      isEditModalOpen: false,
+      editWidgetData: {},
       dashboardFilter: ''
     };
     props.eventEmitter.on('projectChange', () => {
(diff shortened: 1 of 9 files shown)


frontend/src/dashboard.js
```diff
@@ -445,6 +484,14 @@ export class Dashboard extends React.Component {
...
           onClose={this.onDeleteWidgetClose}
         />
+        {this.state.isEditModalOpen ?
+          <EditWidgetModal
+            isOpen={this.state.isEditModalOpen}
+            onSave={this.onEditWidgetSave}
+            onClose={this.onEditWidgetClose}
+            data={this.state.editWidgetData}
+          />
+        : ''}
```
rsnyman, line 494: You don't need the this. Just put the `EditWidgetModal` in the page. The `isEditModalOpen` value determines whether or not it is visible. This way you have less render cycles.

rsnyman (changes requested):

frontend/src/dashboard.js
Griffin-Sullivan: So the reason I'm doing it this way is because we have to update the `EditWidgetModal` with the values we are passing into props as `data`. In the constructor I pull the values out of `props.data` to be the default value in the text inputs. If I just put the `EditWidgetModal` on the page, it will bring up a broken modal with missing param Text Inputs and no values in the fields, because when it's rendered we still haven't selected a widget. 

Are there better ways to do this? If I try to force the value of the input fields to be what's in `props.data` it doesn't let you change the text.

rsnyman: Hrm. [This answer](https://stackoverflow.com/a/53062164) on Stack Overflow might help?

Griffin-Sullivan: That didn't work, but I did find [this](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-controlled-component). That would require me to move every state variable and `onChange` method in `EditWidgetModal` to `dashboard.js` and control the entire component from the parent. Would you prefer that over my current implementation?

rsnyman: I'm wondering how we do it in other places, like the admin pages, let me have a look.

rsnyman (approved):

PR #420, opened 2023-02-01
Add OCP templates for AppInterface

rsnyman:


ocp-templates/app-interface/backend.yaml
```diff
@@ -146,243 +145,41 @@ objects:
...
-- kind: DeploymentConfig
+    host: ${BACKEND_ROUTE}
+    to:
+      kind: Service
+      name: ibutsu-backend
+    tls:
+      insecureEdgeTerminationPolicy: Redirect
+      termination: edge
+# ------------------------------------------------
+- kind: Secret
```
bsquizz, line 166: Just FYI: Most apps would have secret like this inserted inserted into the namespace config via vault but with a special syntax, app-interface does support passing vault creds into the template as parameter values -- so we can make this work.

ocp-templates/app-interface/scheduler.yaml
```diff
@@ -0,0 +1,136 @@
...
+              secretKeyRef:
+                key: db.name
+                name: ibutsu-db
+          - name: REDIS_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                key: db.password
+                name: redis
+                optional: true
+          - name: REDIS_HOSTNAME
```
bsquizz, line 61: Should we be fetching the redis values from a secret in a similar way in the other templates too?

ocp-templates/app-interface/scheduler.yaml
```diff
@@ -0,0 +1,136 @@
...
+  value: ibutsu.example.com
+- name: REDIS_PASSWORD
+  displayName: Redis Password
+  description: The password for Redis
+  generate: expression
+  from: '[\w]{16}'
+- name: REDIS_HOST
+  displayName: Redis Hostname
+  value: redis.svc
+- name: POSTGRESQL_HOST
```
bsquizz, line 131: I don't think you're using this param in your template

ocp-templates/app-interface/scheduler.yaml
```diff
@@ -0,0 +1,136 @@
...
+  value: iqe-ibutsu-server
+- name: BACKEND_ROUTE
+  displayName: Backend Route
+  description: The URL of the backend of the Ibutsu server
+  value: ibutsu-api.example.com
+- name: FRONTEND_ROUTE
+  displayName: Frontend Route
+  description: The URL of the frontend of the Ibutsu server
+  value: ibutsu.example.com
+- name: REDIS_PASSWORD
```
bsquizz, line 123: Same here -- I think you're no longer using this param since you're fetching the value from a secret.

ocp-templates/app-interface/scheduler.yaml
```diff
@@ -0,0 +1,136 @@
...
+  displayName: Redis Password
+  description: The password for Redis
+  generate: expression
+  from: '[\w]{16}'
+- name: REDIS_HOST
```
bsquizz, line 128: Same here -- I think you're no longer using this param since you're fetching the value from a secret.

bsquizz (changes requested):
This looks like a solid start... once you get redis and the DB provisioned via app-sre, their tooling should put a secret into the NS that your configs will be able to consume to read the values of hostname/username/password/etc.

Just make sure you remove the un-used parameter at the bottom if you no longer use params to read in those values.

ocp-templates/app-interface/scheduler.yaml
rsnyman: Probably, I'll fix that up with my changes for redis and rds.

ocp-templates/app-interface/backend.yaml
rsnyman: I don't mind which way we do it, Vault sounds like a good option to me.

ocp-templates/app-interface/scheduler.yaml
rsnyman: Technically I use it for the celery URLs. Which is another question I have.... how do I populate those from the secrets that are injected into the environment?

bsquizz: What if you also read the URL from a secret? For example if there's a secret named `ibutsu-celery` with key `url`. I will show you how we can create a custom secret in app-interface that will fetch the username and password of the provisioned redis instance.

rsnyman: That would work. Thanks!

bsquizz (approved):
The patterns here are following what I commonly see in for other apps in app-interface, looks good! I think at this point we should deploy this and see how it goes.

Issue #411, opened 2022-11-24
[Enhancement] Add dashboard widget for Test Run Heatmap


ntkathole:
Currently ibutsu frontend have widget for Jenkins Pipeline Heatmap (A heatmap of test runs from a Jenkins pipeline). But there are certain scenarios where results are not uploaded via Jenkins or with Jenkins metadata. 

In such cases, we need similar widget to Jenkins Pipeline Heatmap that will collect test runs as per filters irrespective of jenkins job metadata. 

Example Scenario : Stage post deploy tests results which are not directly uploaded to ibutsu and fetched from aws bucket.

rsnyman:
This should beg resolved with #428 so I'm going to close this issue.

PR #428, opened 2023-02-22
New filter-heatmap widget for making flexible heatmaps

Griffin-Sullivan:
Adding a new widget for a heatmap that collects runs based off of a filter. Users can specify a list of filters, number of builds, and a group_field to get a heatmap of runs that fit those parameters. Open to suggestions especially on how I implemented the backend. I sort of had to retrieve all of the most recent run results and then manually filter them to make sure they're in the right order and then only return x amount of runs based on what the user passed for builds. Frontend is pretty much the exact same as jenkins-heatmap.

### DIFF
--- a/backend/ibutsu_server/constants.py
+++ b/backend/ibutsu_server/constants.py
@@ -107,6 +107,34 @@ WIDGET_TYPES = {
         ],
         "type": "widget",
     },
+    "filter-heatmap": {
+        "id": "filter-heatmap",
+        "title": "Filtered Heatmap",
+        "description": "A heatmap of filtered test runs.",
+        "params": [
+            {
+                "name": "filters",
+                "description": "The filters for the runs to be included" " in the query.",
+                "type": "string",
+                "required": True,
+            },
+            {
+                "name": "builds",
+                "description": "The number of builds to analyze.",
+                "type": "integer",
+                "default": 5,
+                "required": True,
+            },
+            {
+                "name": "group_field",
+                "description": "The field in a result to group by, typically 'component'",
+                "type": "string",
+                "required": True,
+                "default": "component",
+            },
+        ],
+        "type": "widget",
+    },
     "run-aggregator": {
         "id": "run-aggregator",
         "title": "Run Aggregation",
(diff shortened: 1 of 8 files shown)


backend/ibutsu_server/constants.py
```diff
@@ -107,6 +107,41 @@
...
+                "required": True,
+                "default": "component",
+            },
+            {
+                "name": "count_skips",
+                "description": "Count skips against the pass rate.",
+                "type": "boolean",
+                "required": False,
+                "default": True,
+            },
```
rsnyman, line 141: Let's drop this, don't count skips against the pass rate.

frontend/src/dashboard.js
```diff
@@ -361,6 +361,16 @@ export class Dashboard extends React.Component {
                         onEditClick={() => this.onEditWidgetClick(widget.id)}
                       />
                     }
+                    {(widget.type === "widget" && widget.widget === "filter-heatmap") &&
+                      <JenkinsHeatmapWidget
```
rsnyman, line 365: We should actually switch these around, convert the `JenkinsHeatmapWidget` to a `FilterHeatmapWidget` and then make the Jenkins heatmap use the `FilterHeatmapWidget`

rsnyman (changes requested):
Overall, this looks good. I'd like to drop the "count skips as fails/passes" since this doesn't actually make any sense, and I'd like to make the Jenkins heatmap a sort of a type of filter heatmap.

I also just want to test this out manually, locally before giving the final go-ahead.

j0nallen:
@Griffin-Sullivan Could this be a solution to [https://github.com/ibutsu/ibutsu-server/issues/411](https://github.com/ibutsu/ibutsu-server/issues/411)?

Griffin-Sullivan:
> @Griffin-Sullivan Could this be a solution to #411?

@j0nallen I believe so! I was unaware of that issue, but it seems like this will cover that use case.

j0nallen:
Thanks, @Griffin-Sullivan - will try this out and see!

PR #444, opened 2023-03-24
Set the run ID if the filename contains a UUID

rsnyman:


backend/ibutsu_server/test/test_importers.py
```diff
@@ -35,7 +35,9 @@ def test_update_import_status(self, mocked_session):
     def test_junit_import(self, mocked_session, mocked_update, MockImport, MockImportFile):
         """Test the junit importer"""
         mocked_import = {"id": "12345"}
-        mocked_record = MagicMock(data={})
+        mocked_record = MagicMock(
+            data={}, filename="run-f56b3831-0813-484a-809f-53d4c36a10a5.ibutsu.xml"
```
mrpetrbartos, line 39: we might want to assert that the run id was really set

rsnyman: Good point.

mrpetrbartos (approved):

PR #451, opened 2023-03-30
Add the ability to disable SSL verification

rsnyman:
WARNING: Setting this to "false" is insecure! Use with extreme caution!


jjaquish (approved):
Not tested locally, but LGTM

PR #454, opened 2023-03-31
Fix up font colour issues with the labels on various charts/widgets

rsnyman:
Before:

image

After:

image


catastrophe-brandon (approved):
I'm no CSS wizard, but that is definitely CSS.

PR #465, opened 2023-06-20
Set socket timeouts for redis

bsquizz:
When disconnected from redis, the worker does not recover and begin consuming tasks again.

https://github.com/celery/celery/issues/8030 suggests adding socket timeout options -- which is what the goal of this PR is.


bsquizz:
Tested with docker compose dev env and restarting the redis container. The celery worker is at least running. Let's see how this operates in "the real world" when the connection to AWS Elasticache gets flaky.

PR #467, opened 2023-07-27
create-admin password AttributeError work-around

catastrophe-brandon:
Using the current hybrid_property setter, an AttributeError surfaces when using the --create-admin functionality. The current code looks correct when compared against the SQLAlchemy documentation, so I've implemented a light-weight work-around. It involves setting the password without using the hybrid property setter. I suspect a true solution would involve an upgrade of SQLAlchemy, which would be a heavy lift for fixing such simple functionality.

### DIFF
--- a/backend/ibutsu_server/db/models.py
+++ b/backend/ibutsu_server/db/models.py
@@ -238,11 +238,18 @@ class User(Model, ModelMixin):
         return self._password
 
     @password.setter
-    def password_set(self, value):
+    def password_set(self, value: str):
         self._password = bcrypt.generate_password_hash(value).decode("utf8")
 
     def check_password(self, plaintext):
         return bcrypt.check_password_hash(self.password, plaintext)
+
+    def set_password(self, value: str):
+        """
+        The password setter breaks with an AttributeError, so this is a temporary work-around
+        until that can be resolved.
+        """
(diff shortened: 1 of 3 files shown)


backend/setup.py
```diff
@@ -20,7 +20,7 @@
     "gunicorn",
     "kombu",
     "lxml",
-    "psycopg2",
+    "psycopg2-binary",
```
catastrophe-brandon, line 23: This probably isn't necessary for the PR; I mostly did this to make my life easier.

rsnyman: Yeah, in order to get Ibutsu to run in FedRAMP, the psycopg2 module has to be built from source, so this needs to be removed.

backend/ibutsu_server/db/models.py
```diff
@@ -238,12 +238,19 @@ def password(self):
         return self._password
 
     @password.setter
-    def password_set(self, value):
+    def password_set(self, value: str):
```
rsnyman, line 241: You might want to make this property use the `set_password` method below, so that the password hash generation happens in only one place. This will prevent future frustration when someone accidentally only changes one but not the other.

LightOfHeaven1994:
Already took care of it here https://github.com/ibutsu/ibutsu-server/pull/471

PR #469, opened 2023-09-08
Update filtering of results to include local tests runs

patriksuba:
Fixing issue mentioned in [IQE-2249](https://issues.redhat.com/browse/IQE-2249).

### DIFF
--- a/backend/ibutsu_server/controllers/result_controller.py
+++ b/backend/ibutsu_server/controllers/result_controller.py
@@ -1,10 +1,12 @@
 from datetime import datetime
+from sqlalchemy import or_
 
 import connexion
 from ibutsu_server.db.base import session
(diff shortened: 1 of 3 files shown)


mshriver (changes requested):
Discussed at some length in chat, I don't think this is the way we should approach resolving the original issue.

This changes the default behavior when including a jenkins job name filter, removing the implicit filtering that a job name is set at all.   I think for UI ibutsu users, this will include results that they don't expect when defining a jenkins job name filter.

I suggest we actually make changes to the iqe-requirements IbutsuClient class, and construct the explicit filters that we want to get results for instead of relying on implicit behaviors of ibutsu-server.

patriksuba:
I agree with Mike. I will create a solution in iqe-requirements IbutsuClient class.

PR #471, opened 2023-11-24
Fix local dev setup

LightOfHeaven1994:


RonnyPfannschmidt (approved):

backend/setup.py
```diff
@@ -6,7 +6,7 @@
 REQUIRES = [
     "alembic",
     "celery",
-    "connexion",
+    "connexion[flask]",
```
rsnyman, line 9:
```suggestion
    "connexion[swagger-ui]==2.14.2",
```

docker-compose.dev.yml
```diff
@@ -32,13 +33,20 @@ services:
...
       - "8080:8080"
     depends_on:
-      - postgres
-      - redis
+      postgres:
+        condition: service_healthy
+      redis:
+        condition: service_started
+    healthcheck:
+      test: curl --output /dev/null --silent --head --fail http://localhost:8080
```
rsnyman, line 46: This should probably be something like `/api/health` -- there is a health endpoint.

LightOfHeaven1994: Fixed

docker-compose.dev.yml
```diff
@@ -54,17 +62,25 @@ services:
...
+        condition: service_started
+      backend:
+        condition: service_healthy
   postgres:
     environment:
       POSTGRES_USER: ibutsu
       POSTGRES_PASSWORD: ibutsu
       POSTGRES_DB: ibutsu
-    image: postgres:14
+    image: postgres:latest
```
rsnyman, line 79: You don't want to do this. PostgreSQL versions are not compatible with each other, you need to do a manual data migration. If someone is using postgres 14, and then they suddenly get postgres 15, it won't start because it'll detect that the data is in a previous format.

You are welcome to use postgres:15 instead of 14, but don't change to "latest" because that'll cause issues.

LightOfHeaven1994:
@rsnyman feel free to merge it when you have time

rsnyman (changes requested):
Everything looks fine except the postgres change, which I've commented on.

docker-compose.dev.yml
rsnyman: Incidentally, I learned this the hard way :-(

LightOfHeaven1994: @rsnyman didn't think about it since it's local dev setup only. Good catch, fixed

rsnyman (approved):

PR #473, opened 2023-11-30
Make default dashboard

LightOfHeaven1994:


rsnyman (approved):
Looks good to me.

My only caveat is that in JS you use the `!!` to convert a value to a boolean. So while this probably still works, you may run into funny issues in the future. JS's truthiness can be a little quirky.

LightOfHeaven1994:
> Looks good to me.
> 
> My only caveat is that in JS you use the `!!` to convert a value to a boolean. So while this probably still works, you may run into funny issues in the future. JS's truthiness can be a little quirky.

aha, didn't know about it. Will return it back in next commit.
I need to fix other small thing as default widgets work now, I wish Select also has value set instead of being empty. I will put gif when finish with it "before" & "after".

Thanks for the review btw :+1:

PR #476, opened 2024-02-13
Add mpp templates

Victoremepunto:
This PR aims to:
- deprecate `app-inteface` set of templates
- Add new set of updated templates (`mpp`) to be compliant to OCP > 4.14
  - Updates to use `Deployment` over `DeploymentConfig`
  - Updates `CronJob` specs
- Small fixes and tweaks


ocp-templates/README.md
```diff
@@ -16,6 +18,9 @@ The prod templates are set up to use the images at [quay.io/organization/ibutsu]
 ``latest``. These images are built and tagged as ``latest`` on every release of Ibutsu. These are meant to deploy the production instance of Ibutsu. The templates include some extra things like database
 backups and vacuums.
 
+## Prod
```
digitronik, line 21: ^^

ocp-templates/README.md
```diff
@@ -16,6 +18,9 @@ The prod templates are set up to use the images at [quay.io/organization/ibutsu]
...
+The MPP templates are meant to be used on an OCPv4 Openshift cluster with multi-tennant support.
```
digitronik, line 22: I think better not to use specific cluster `multi-tenant` arch. It will useful to anyone where they want to go with `keycloak` signing.

ocp-templates/mpp/backend.yaml
```diff
@@ -7,23 +7,24 @@ metadata:
...
-  apiVersion: v1
+- kind: Deployment
+  apiVersion: apps/v1
   metadata:
-    labels:
-      app: ${APP_NAME}
     name: ibutsu-backend
+    namespace: ${NAMESPACE}
+    annotations:
+      [{"from":{"kind":"ImageStreamTag","name":"ibutsu-backend:latest","namespace":"${NAMESPACE}"},"fieldPath":"spec.template.spec.containers[?(@.name==\"container\")].image","pause":"false"}]
```
digitronik, line 16: do we need this?

ocp-templates/mpp/backend.yaml
```diff
@@ -147,48 +118,101 @@ objects:
...
+    tags:
+    - from:
+        kind: DockerImage
+        name: quay.io/ibutsu/backend:latest
+      importPolicy:
+        scheduled: true
+      name: latest
+    - from:
+        kind: DockerImage
+        name: quay.io/ibutsu/backend:master
```
digitronik, line 211: do we need this in prod?

ocp-templates/mpp/frontend.yaml
```diff
@@ -1,4 +1,4 @@
-# Template for Ibutsu frontend in AppInterface
+# Template for ibutsu prod server
```
digitronik, line 1: frontend ^^

ocp-templates/mpp/frontend.yaml
```diff
@@ -7,32 +7,33 @@ metadata:
...
 # ===============================================
 objects:
-- kind: DeploymentConfig
-  apiVersion: v1
+- kind: Deployment
+  apiVersion: apps/v1
   metadata:
     labels:
-      app: ${APP_NAME}
+      app: ibutsu-frontend
```
digitronik, line 14: I would use same app name every where.  If we want separate labeling for frontend and backend we can add that as well.

ocp-templates/mpp/tenant-namespace.yml
```diff
@@ -0,0 +1,37 @@
+apiVersion: template.openshift.io/v1
```
digitronik, line 1: Do we you we need to ship this? 
My personal thought, Its part of admin stuff.  Secondly its common for all users of that specific cluster. 
I want to say cluster specific task.

digitronik (changes requested):
Some small requests else LGTM

ocp-templates/README.md
Victoremepunto: agreed

ocp-templates/mpp/backend.yaml
Victoremepunto: nope, removed.

ocp-templates/mpp/frontend.yaml
Victoremepunto: thanks, fixed

ocp-templates/mpp/frontend.yaml
Victoremepunto: For Deployment objects, the label is used for the match pod policy, so in this case, every pod requires different labels.

I agree though, that for consistency, we can have the same set of labels for the same app, but I don't feel it should be part of this PR and more like part of a followup. I don't want to change labels across all objects right now, but happy to do in a followup

ocp-templates/mpp/tenant-namespace.yml
Victoremepunto: removed from the PR

ocp-templates/mpp/backend.yaml
Victoremepunto: Short answer: yes

The annotations are required for wiring up the Deployment (K8s object) with ImageStreams (Openshift objects). There's [official Openshift docs around this](https://docs.openshift.com/container-platform/4.14/openshift_images/triggering-updates-on-imagestream-changes.html#images-triggering-updates-imagestream-changes-kubernetes-about_triggering-updates-on-imagestream-changes)

Feel free to reach out for a extended version, sorry its not straightforward to explain :)

Victoremepunto: I actually had to add it to the rest of the Deployment objects.

LightOfHeaven1994 (approved):
Looks good to me, we went through templates on a meeting and made it work correctly. Thank you, Victor!

LightOfHeaven1994:
Failure is not related to this PR, tested with dummy PR https://github.com/ibutsu/ibutsu-server/actions/runs/7901012519/job/21563774045?pr=477 , we are good to merge it

PR #478, opened 2024-02-14
Update OCP mpp backend templates

patchkez:
- we need to add https:// protocol to the FRONTEND_URL/BACKEND_URL values, otherwise the SSO auth will be failing with an invalid redirect-uri error
- JWT secret + var must be used in the backend pod, without it we get error `ibutsu_server.errors.IbutsuError: JWT_SECRET is not defined in configuration or an environment variable`


LightOfHeaven1994 (approved):
lgtm, good catch!

PR #479, opened 2024-02-15
Add resolution to victory core package

LightOfHeaven1994:
Resolves react-chart dependency issue. For more info see https://github.com/patternfly/patternfly-react/issues/10064

Note that PR has both `resolutions` (for yarn) and `overrides` (for npm). We might need to take a look later how to deal with it and use only one package manager because our tests use yarn and I see yarn is mentioned in the doc but npm is used in Dockerfile to build FE

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -43,6 +43,12 @@
     "typescript": "^4.9.5",
     "wolfy87-eventemitter": "^5.2.9"
   },
+  "resolutions": {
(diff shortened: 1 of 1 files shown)


Victoremepunto (approved):
This looks like working properly, nice work !

Please, lets card up to use `yarn` or `npm` across the board, for consistency, do this on a folloupw

mshriver (approved):

PR #482, opened 2024-02-19
Add route for CNAME records for frontend and backend services

patchkez:


ocp-templates/mpp/backend.yaml
```diff
@@ -174,6 +174,24 @@ objects:
...
+    name: ibutsu-backend
+    namespace: ${NAMESPACE}
+  spec:
+    host: ${BACKEND_ROUTE_CNAME}
+    tls:
+      insecureEdgeTerminationPolicy: Redirect
+      termination: edge
+    to:
+      kind: Service
+      name: ibutsu-backend
```
Victoremepunto, line 194: why are we creating 2 routes ? shouldn't we just have the first one ? (the one that's used by the CNAME ?)

ocp-templates/mpp/backend.yaml
```diff
@@ -25,7 +25,7 @@ objects:
     template:
       metadata:
         labels:
-          app: ibutsu-backend
+          app: ${APP_NAME}
```
Victoremepunto, line 28: This is going to break the Service if the APP_NAME is not set correctly, the APP_NAME's default value is `ibutsu-server` - which is also used by all the other templates (with the same value), while the Service's label selector should use a unique value (otherwise more than one pod will match "app=ibutsu-server"

ocp-templates/mpp/frontend.yaml
```diff
@@ -144,6 +163,10 @@ parameters:
   displayName: Frontend Route
   description: The URL of the frontend of the Ibutsu server
   value: ibutsu.example.com
+- name: FRONTEND_ROUTE_CNAME
+  displayName: Frontend Route
+  description: The CNAME URL of the frontend of the Ibutsu server
+  value: ibutsu.prod.com
```
Victoremepunto, line 169: Why do we need separate route parameters ? can't we use just 1 route, set the value at creation time with the CNAME URL ?

Victoremepunto (changes requested):
I understand what this PR is trying to do, but I don't understand why we need to add additional routes.

There's a potential breakage of the Service if the label is used with the default values too, I think this needs a small review

PR #498, opened 2024-05-31
BROKEN LINT: ruff lint/format implementations

mshriver:
This fails the lint config, I need to apply updates to resolve


LightOfHeaven1994 (approved):
LGTM. 

Note: I'm thinking that we need to have pinned packages instead of what we are doing now. As we had some issues in the past because packages randomly updated. This is out of scope of this PR, but we will need to do that to ensure that ibutsu will not break on prod because of some random package

backend/docker/Dockerfile.fuzz_testing
```diff
@@ -1,10 +1,14 @@
+# hadolint global ignore=DL3008,DL3013,DL3041
 FROM python:3.9
```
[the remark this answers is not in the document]
mshriver: Probably - I'll rebase on your react PR and take a second look.

.ruff.toml
```diff
@@ -0,0 +1,7 @@
+line-length = 100
+
+[lint]
+select = ["F", "I", "UP"]
```
mshriver, line 4: @jjaquish @LightOfHeaven1994 @RonnyPfannschmidt any other ruff linters that you'd like to include at this time?

LightOfHeaven1994: @mshriver these look good to me :+1:

podman_build.sh
```diff
@@ -1,22 +1,30 @@
...
-    IMAGES_TO_BUILD=("backend", "scheduler", "worker", "flower", "frontend")
+    IMAGES_TO_BUILD=("backend" "scheduler" "worker" "flower" "frontend")
 else
     IMAGES_TO_BUILD=("$1")
 fi
 
 # Image info
 IMAGE_TAG=$(git rev-parse --short=7 HEAD)
-IMAGE_PREFIX="quay.io/cloudservices/ibutsu-"
+IMAGE_PREFIX="ibutsu/"
```
LightOfHeaven1994, line 11: I assume this script has been made for cases when we want to build image locally and push it to quay registry. So I would keep correct prefix - `IMAGE_PREFIX="quay.io/ibutsu/"`

LightOfHeaven1994 (approved):
Tested it locally and seems everything works fine. Only one suggestion added

podman_build.sh
mshriver: yeah, I think I'm going to move this to a flag or arg, because I'd like the default here to just build local images aligned with the image names in ibutsu_pod.yml and/or the docker-compose file.

That way it can be used to build the images locally without tagging them with the upstream destination.

I can set the default to be quay.io tagged, so when building just for local use the flag gets passed?

podman_build.sh
```diff
@@ -1,22 +1,58 @@
 #!/bin/bash
 
-if [[ "$1" == "" ]]; then
-    IMAGES_TO_BUILD=("backend", "scheduler", "worker", "flower", "frontend")
-else
+function print_usage() {
```
mshriver, line 3: @LightOfHeaven1994 significant changes here if you want to re-review

.pre-commit-config.yaml
```diff
@@ -12,17 +9,29 @@ repos:
...
-  - id: flake8
-    args: [--max-line-length, "100", --ignore, "E203, W503"]
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.15.2
+    - id: ruff
+      args:
+        - '--fix'
+    - id: ruff-format
+
+## TODO: enable
```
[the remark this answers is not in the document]
mshriver: yeah this hook will introduce a TON of changes and failures that need to be resolved. I left it here so it doesn't get completely ignored forever.

PR #499, opened 2024-05-31
Update from localhost to 127.0.0.1

mshriver:
Having issues in podman 5 with pasta networking defaults, using localhost to resolve other containers in the pod.

Hit this moving to podman 5 with Fedora 40. I'm actually using podman without docker, and have even added a podman.sock alias trying to resolve this.

I don't think changing to 127.0.0.1 in these places is detrimental, where using localhost will be dependent on the podman version, and whether rootful network is being used. 

This manifests in three places:

### backend 

Define constant and import it for references to `localhost`, changing them to 127.0.0.1. These are only used as a default, and won't be relevant in production.

### frontend

An error will occur in login.json and no login options will render - similar to the issue reported for SSO integration when CA certs are not setup for the client.  There is a close connection here to the behavior of the login script that needs specific attention, but that should be in a separate PR. This only addresses the default serverUrl in local environments.

Again this should not impact production, where serverUrl is set. 
``` 
componentDidMount() {
    HttpClient.get([Settings.serverUrl, 'login', 'support'])
    
 ```
 
 
 ### Scripts

Scripts for local development pod env deployment also reference localhost, so I have updated references here too. These again do not impact production deployment to openshift.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,11 @@
 repos:
-- repo: https://github.com/asottile/reorder_python_imports
-  rev: v3.9.0
-  hooks:
-  - id: reorder-python-imports
 - repo: https://github.com/ambv/black
-  rev: 22.10.0
+  rev: 24.4.2
   hooks:
   - id: black
     args: [--safe, --quiet, --line-length, "100"]
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.3.0
+  rev: v4.6.0
   hooks:
   - id: trailing-whitespace
   - id: end-of-file-fixer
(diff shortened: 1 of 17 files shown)


jjaquish:
I'm not necessarily opposed to this one but it seems like an application fix for a system issue. I'm wondering if we should spend some more time looking into the environment issues first?

If this ends up getting merged in we might also want to add some comments for swapping this back later.

mshriver:
@jjaquish agreed - felt wrong but I dug fairly deep on this and I think moving forward, with podman, it's going to take non-default config and legacy network use to keep using localhost.

Are you using podman locally? Almost everyone I've checked with is using docker / docker-compose to run locally, or older podman. Podman 5 moved to a new network stack. 

I updated the description to explain a bit more in detail. Since I think the impact/risk will be low, and only to dev environments, I'm going to merge this. We can be quick to revert if need be.

PR #500, opened 2024-06-04
DRAFT: Satellite dashboard

mshriver:
Very very early draft, in design and discussion at the moment.

The goal is to deliver a static-ish dashboard, similar to iqe-requirements, but for cross-project test result aggregation.

Currently struggling with a split menu toggle + dropdown implmentation, similar to insights:
https://github.com/RedHatInsights/frontend-components/blob/69399c1ff2e3be6fd8e387ce768a6435efcd851d/packages/components/src/ConditionalFilter/ConditionalFilter.tsx#L185

commented out dual-select is easy and working, but there was review input that they look dependent on each other. I'm skeptical of this input, as their selections will be completely different, and will completely change the view. There will still be dashboards under the projects, and those dashboards won't be visible/searchable without selecting the project first. 

This is built on the commits for #498  and #499

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -23,7 +23,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python-version: [ "3.8", "3.9", "3.10", ] # five is broken on 3.11 - celery update needed "3.11.0-alpha - 3.11"]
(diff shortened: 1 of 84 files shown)


mshriver:
Commits from here split into several other PRs, going to close this out.

PR #504, opened 2024-06-17
Portal Dashboards, PF5

mshriver:


backend/ibutsu_server/controllers/admin/portal_controller.py
```diff
@@ -0,0 +1,153 @@
...
+
+    :rtype: Portal
+    """
+    check_user_is_admin(user)
+    portal = Portal.query.get(id_)
+    if not portal:
+        portal = Portal.query.filter(Portal.name == id_).first()
+    if not portal:
+        abort(HTTPStatus.NOT_FOUND)
+    return portal.to_dict(with_owner=True)
```
vcwild, line 53: To avoid DRY maybe we can use:
```suggestion
    if portal := Portal.query.get(id_) or Portal.query.filter(Portal.name == id_).first():
        return portal.to_dict(with_owner=True)
    abort(HTTPStatus.NOT_FOUND)
```

mshriver: Thanks, I'll take a look at improving there! I copied some patterns that looked unnecessarily verbose and it would be great to condense them here and in the projects controller.

vcwild: Ok. I hate walrus but it's actually useful sometimes

mshriver: I'm learning to like it, and it certainly does make these blocks cleaner.

backend/ibutsu_server/controllers/admin/portal_controller.py
```diff
@@ -0,0 +1,151 @@
...
+
+    :rtype: Portal
+    """
+    check_user_is_admin(user)
+
+    # get by ID or check if the portal name matches the passed ID
+    if portal := Portal.query.get(id_) or Portal.query.filter(Portal.name == id_).first():
+        return portal.to_dict(with_owner=True)
+    else:
+        abort(HTTPStatus.NOT_FOUND)
```
vcwild, line 52: I think maybe we don't need else here

backend/ibutsu_server/controllers/admin/portal_controller.py
```diff
@@ -0,0 +1,151 @@
...
+        # Grab the fields from the request
+        portal_dict = connexion.request.get_json()
+
+        # If the "owner" field is set, ignore it
+        portal_dict.pop("owner", None)
+
+        # update the portal info
+        portal.update(portal_dict)
+        session.add(portal)
+        session.commit()
```
vcwild, line 134: Maybe session should be abstracted away in a service object and have all the assertion predicates correct before trying to commit the update to the db?

backend/ibutsu_server/controllers/widget_config_controller.py
```diff
@@ -23,22 +24,46 @@ def add_widget_config(widget_config=None, token_info=None, user=None):
...
+
+    # bad request checks
     if data["widget"] not in WIDGET_TYPES.keys():
         return "Bad request, widget type does not exist", HTTPStatus.BAD_REQUEST
+    # TODO: openapi plugins to support exclusive options at the schema level
+    if (
+        not any(data.get(key) for key in ["portal_id", "portal", "project_id", "project"])
+        or (data.get("project_id") or data.get("project"))
+        and (data.get("portal_id") or data.get("portal"))
+    ):
```
vcwild, line 36: Maybe serialize this via pydantic or a dataclass to ensure data consistency?

vcwild (changes requested):
I left a few comments, otherwise solid piece of work with the Router and FE Components.

backend/ibutsu_server/openapi/openapi.yaml
```diff
@@ -2388,6 +2661,33 @@ components:
           description: The date this artifact was uploaded
           type: string
       type: object
+    Portal:
+      example:
+        id: fe5ad2dc-330a-11ef-9036-12b95372ee33
+        name: my-portal
+        title: My portal
+        owner_id: 07a776ba-330b-11ef-9338-12b95372ee33
+      properties:
```
RonnyPfannschmidt, line 2670: this one is missing the required entry

backend/ibutsu_server/controllers/admin/portal_controller.py
mshriver: correct, I like the explicit behavior of this, but I think it's ultimately subjective. definitely looks cleaner without it.

backend/ibutsu_server/controllers/admin/portal_controller.py
mshriver: I do like this idea, but for the moment I'm going to say I'd like to keep it out of the scope of this PR? And then we can update all of the controllers with a service object for the session and common assertions happening implicitly.

backend/ibutsu_server/controllers/widget_config_controller.py
mshriver: Taken together with processing the project/portal in order to set the respective ID, I think this would combine a few logic blocks.

The model class itself may suffice, I'll investigate this further.

Thanks!

backend/ibutsu_server/openapi/openapi.yaml
mshriver: agreed - as are all the other schema here. Not sure why, conceptually there are absolutely required fields in these objects.

Will take a second look, might be an easy API spec "upgrade".

RonnyPfannschmidt: it may be a god idea to do a typespec import/fixup as a followup

frontend/src/components/ibutsu-header.js
```diff
@@ -122,10 +114,33 @@ export class IbutsuHeader extends React.Component {
...
+    const {setPrimaryObject, primaryId} = this.context;
+
     // adding s here seems risky, but this scope is small, it's only abstracted for 2 things
     HttpClient.get([Settings.serverUrl, endpoint], params)
       .then(response => HttpClient.handleResponse(response))
-      .then(data => this.setState({projects: data[endpoint+'s'], filteredProjects: data[endpoint+'s']}));
+      .then(data => {
+        // populate context and current input value if passed
+        for ( let item of data[endpoint+'s'] ) {
+          if (item.id === primaryId) {
```
Andrewgdewar, line 126: is this likely to be one value? 
as seeing a setState in a loop hurts. 

I sugggest 

 const itemYourAfter =  Object.keys(data[endpoint+'s']).find(key=>data[endpoint+'s'][key].id === primaryId)
 
 ...whatevs

Andrewgdewar: data[endpoint+'s']?.[key]?.id   << this may be necessary

ColeHiggins2 (commented):
Looks good so far! These are no pending comments. Just trying to learn!

backend/ibutsu_server/controllers/portal_controller.py
```diff
@@ -0,0 +1,129 @@
...
+    return {
+        "portals": [portal.to_dict() for portal in portals],
+        "pagination": {
+            "page": page,
+            "pageSize": page_size,
+            "totalItems": total_items,
+            "totalPages": total_pages,
+        },
+    }
+
```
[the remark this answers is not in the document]
mshriver: correct, it's like deleting a project.

backend/ibutsu_server/controllers/portal_controller.py
```diff
@@ -0,0 +1,129 @@
...
+@validate_uuid
+def get_portal(id_, token_info=None, user=None) -> dict:
+    """Get a single portal by ID
+
+    :param id: ID of test portal
+    :type id: str
+
+    :rtype: Portal
+    """
+    if not is_uuid(id_):
```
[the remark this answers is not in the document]
mshriver: Will double check, at this point I was repeating a lot of the patterns from the project controller because the entities should have the same CRUD behaviors and permissions.

backend/ibutsu_server/test/test_portal_controller.py
```diff
@@ -0,0 +1,155 @@
...
+    id=MOCK_ID,
+    name="unittest-portal",
+    title="UnitTest Portal",
+    owner_id=MOCK_USER_ID,
+    default_dashboard_id="7e8a1684-3306-11ef-8433-12b95372ee33",
+)
+MOCK_USER = MockUser.from_dict(**{"id": MOCK_USER_ID})
+
+
+class TestPortalController(BaseTestCase):
```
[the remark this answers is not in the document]
mshriver: yes and yes!

The tests for admin controllers are lacking ... I'll try to add them in this scope.

frontend/src/components/ibutsu-header.js
mshriver: got it! I can see how that's way better.

mshriver:
```
        const paramMatch = Object.keys(data[endpoint+'s']).find(key=>data[endpoint+'s'][key].id === primaryObject.id)
        setPrimaryObject(paramMatch || null);
        this.setState(
          {
            projects: param_match?.title || data[endpoint+'s'],
            filteredProjects: param_match ? '' : data[endpoint+'s'],

          })
      }
```

Got this down to the above, thanks for the tip about using `find` here instead of looping. It also prompted me to use primaryObject instead of the param, which needed to change since the param should be synced once to the context.

mshriver:
Thanks all for first round reviews - I've re-requested reviews as I've made significant frontend updates that should be ready for local testing and review. 

I'd really appreciate your help in running this locally and finding any weird behaviors! I'll be doing that myself this week, as well as preparing to deploy the stage instance in GPC.

frontend/src/dashboard.js
```diff
@@ -143,23 +198,35 @@ export class Dashboard extends React.Component {
...
+
+    // does it really matter whether I read from params or the context here?
+    // they should be the same, reading from params 'feels' better
+    this.props.navigate('/project/' + this.props.params?.project_id + '/dashboard/' + value?.id)
   };
 
   onDashboardClear = () => {
-    localStorage.removeItem('dashboard');
+    const { setActiveDashboard } = this.context;
+    console.log('clearing dashboard selection');
```
vcwild, line 219: Console log statement

frontend/src/pages/admin/portal-edit.js
```diff
@@ -0,0 +1,483 @@
...
+
+  onSubmitClick = () => {
+    const { portal, selectedOwner, selectedDashboard } = this.state;
+    portal.owner_id = selectedOwner ? selectedOwner.id : null;
+    portal.default_dashboard_id = selectedDashboard ? selectedDashboard.id : null;
+    delete portal.owner;
+    // delete portal.defaultDashboard;
+    this.savePortal(portal.id || null, portal)
+      .then(() => this.props.navigate(-1))
+      .catch((error) => console.error(error));
```
vcwild, line 99: Console statement

frontend/src/pages/admin/portal-edit.js
```diff
@@ -0,0 +1,483 @@
...
+        response = HttpClient.handleResponse(response, 'response');
+        return response.json();
+      })
+      .then(portal => {
+        this.setState({portal: portal, selectedOwner: portal.owner,
+                       inputValueOwner: portal.owner?.name,
+                       selectedDashboard: portal.defaultDashboard,
+                       inputValueDashboard: portal.defaultDashboard?.title});
+      })
+      .catch(error => console.error(error));
```
vcwild, line 144: Console statement

frontend/src/pages/admin/portal-edit.js
```diff
@@ -0,0 +1,483 @@
...
+  }
+
+  getUsers() {
+    HttpClient.get([Settings.serverUrl, 'admin', 'user'])
+      .then(response => {
+        response = HttpClient.handleResponse(response, 'response');
+        return response.json();
+      })
+      .then(data => this.setState({users: data.users, filteredUsers: data.users}))
+      .catch(error => console.error(error));
```
vcwild, line 193: Console statement

frontend/src/pages/admin/portal-list.js
```diff
@@ -0,0 +1,258 @@
...
+        rows: data.portals.map((portal) => portalToRow(portal, this.onDeleteClick)),
+        portals: data.portals,
+        page: data.pagination.page,
+        pageSize: data.pagination.pageSize,
+        totalItems: data.pagination.totalItems,
+        totalPages: data.pagination.totalPages,
+        isEmpty: data.pagination.totalItems === 0
+      }))
+      .catch((error) => {
+        console.error('Error fetching portals data:', error);
```
vcwild, line 136: Console statement

frontend/src/pages/admin/portal-edit.js
mshriver: ack - these errors are intentionally console logged right now (at least they are in the project components).

I'll look at better handling them so we have them in pod logs too?

PR #505, opened 2024-06-17
Use httpstatus module

mshriver:
Use the module imports instead of magic numbers. They probably won't change but some linting/formatting/mypy rules start to enforce it.

These were all the uses of the status codes that I found throughout the backend codebase. There may be some others that are hiding but it's not critical to get 100% of them in this PR.

### DIFF
--- a/backend/ibutsu_server/__init__.py
+++ b/backend/ibutsu_server/__init__.py
@@ -1,4 +1,5 @@
 import os
+from http import HTTPStatus
 from importlib import import_module
 from pathlib import Path
 from typing import Any, Optional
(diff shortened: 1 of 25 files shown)


backend/ibutsu_server/__init__.py
```diff
@@ -137,31 +138,31 @@ def run_task():
         # get params
         params = request.get_json(force=True, silent=True)
         if not params:
-            return "Bad request", 400
+            return HTTPStatus.BAD_REQUEST.phrase, HTTPStatus.BAD_REQUEST
         # get user info
         token = params.get("token")
         if not token:
-            return "Unauthorized", 401
+            return ("Unauthorized",)
```
mshriver, line 145: Fixing

digitronik (approved):
LGTM

backend/ibutsu_server/controllers/import_controller.py
```diff
@@ -89,4 +90,4 @@ def add_import(
         run_archive_import.delay(new_import.to_dict())
     else:
         return "Unsupported Media Type", 415
-    return new_import.to_dict(), 202
+    return new_import.to_dict(), 2
```
Griffin-Sullivan, line 93:
```suggestion
    return new_import.to_dict(), HTTPStatus.ACCEPTED
```
If we change this line shouldn't it be with HTTPStatus?

Griffin-Sullivan (approved):

backend/ibutsu_server/controllers/import_controller.py
mshriver: great eye and catch! This is definitely a typo or bad find+replace entry.

PR #508, opened 2024-08-01
React Context and Project Nested Routing

mshriver:
This pulls the commits from #504 that implement react context and update the react-routing tree.

This introduces some new components for admin and profile pages, splits sidebar into a new component, and refactors many components to use a react context instead of local storage.

This also provides static routing for projects, dashboards, and updated paths for runs and results that are project scoped.

### DIFF
--- a/frontend/src/app.js
+++ b/frontend/src/app.js
@@ -1,13 +1,9 @@
 import React from 'react';
 
-import {
-  Nav,
-  NavList
-} from '@patternfly/react-core';
 import EventEmitter from 'wolfy87-eventemitter';
 import ElementWrapper from './components/elementWrapper';
 
-import { NavLink, Route, Routes } from 'react-router-dom';
+import { Route, Routes } from 'react-router-dom';
 
 import { Dashboard } from './dashboard';
 import { ReportBuilder } from './report-builder';
@@ -15,14 +11,13 @@ import { RunList } from './run-list';
 import { Run } from './run';
 import { ResultList } from './result-list';
 import { Result } from './result';
-import { Settings } from './settings';
 import { View, IbutsuPage } from './components';
-import { HttpClient } from './services/http';
-import { getActiveProject } from './utilities';
+import { IbutsuContext } from './services/context';
+
 import './app.css';
 
-
 export class App extends React.Component {
+  static contextType = IbutsuContext;
   constructor(props) {
     super(props);
     this.eventEmitter = new EventEmitter();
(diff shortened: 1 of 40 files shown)


frontend/src/dashboard.js
```diff
@@ -46,67 +46,84 @@ import {
...
-      localStorage.setItem('dashboard', dashboard);
-      return project.defaultDashboard;
+  sync_context = () => {
+    // Active dashboard
+    const { activeDashboard } = this.context;
+    const { selectedDashboard } = this.state;
+    const paramDash = this.props.params?.dashboard_id;
+    let updatedDash = undefined;
+    // API call to update context
+    if ( paramDash && activeDashboard?.id !== paramDash) {
```
LightOfHeaven1994, line 88: I was able to catch an issue here when I run a fresh ibutsu locally and had only projects created. `const paramDash = this.props.params?.dashboard_id;` returned me a null but as a string. So if didn't catch this and this is what I got:

image

frontend/src/dashboard.js
```diff
@@ -328,7 +393,7 @@ export class Dashboard extends React.Component {
             onChange={this.onTextInputChange}
             id="typeahead-select-input"
             autoComplete="off"
-            placeholder={dashboard ? dashboard.title : "No active dashboard"}
+            placeholder={activeDashboard?.title || "Getting default dashboard..."}
```
LightOfHeaven1994, line 396: I think we need to leave here `"No active dashboard"` placeholder as it clearly says that if there is no dashboard selected and no input value - it means we don't have dashboard selected.

frontend/src/dashboard.js
```diff
@@ -116,23 +133,59 @@ export class Dashboard extends React.Component {
...
+    if ( !activeDashboard && targetObject?.defaultDashboard ){
+      setActiveDashboard(targetObject.defaultDashboard);
+      this.setState({
+        'selectedDashboard': targetObject.defaultDashboard,
+        'dashboardInputValue': targetObject.defaultDashboard?.title
+      })
+    } else {
+      this.setState({
+        'selectedDashboard': 'Select a dashboard',
+        'dashboardInputValue': 'Select a dashboard'
```
LightOfHeaven1994, line 164: See my comment about placeholder (comment on same file). there should be no Input default value as it looks a bit odd. Instead of this we use placeholder. 
Right now it looks a bit odd as something is selected there but I have nothing:
image

frontend/src/utilities.js
```diff
@@ -32,6 +32,8 @@ import {
 import { ClassificationDropdown } from './components';
 
 
+
+
```
LightOfHeaven1994, line 36: too many spaces :smiley:

LightOfHeaven1994 (commented):
Generally looks good, I love context usage and I see some component were converted to functional. There are some things I found and mentioned in my comments. Plus I see some styling issues, I will let you know about them on slack

frontend/src/app.js
```diff
@@ -33,79 +29,71 @@ export class App extends React.Component {
...
+            path="reports"
+            element={<ElementWrapper routeElement={ReportBuilder} eventEmitter={this.eventEmitter} />}
+          />
+
+          <Route
+            path="view/:view_id"
+            element={<ElementWrapper routeElement={View} />}
+          />
+
+          {/* <Route path="*" element={<Navigate to="" replace />} /> */}
```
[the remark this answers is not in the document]
mshriver: I think I want to uncomment this to provide a diaper for bad URLs to redirect. There's one in base.js to redirect to `/project`, I had issues with this one while getting all the nested routing setup correctly.

I'll uncomment it and test with it now, in this case it will also go to `/project` but because of thee routing I think the `to` prop will be empty

frontend/src/components/admin-page.js
```diff
@@ -0,0 +1,59 @@
...
+
+import { IbutsuHeader } from './ibutsu-header';
+import PropTypes from 'prop-types';
+
+
+
+const AdminPage = (props) => {
+    // TODO useEffect instead of eventEmitter prop
+    // TODO notifications on admin page with state and AlertGroup
+    PropTypes
```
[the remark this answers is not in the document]
mshriver: wow good eye!

frontend/src/app.js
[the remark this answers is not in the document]
mshriver: Played with this but couldn't get it to behave like I wanted anyway.

I added catch statements to the HTTP requests to at least stop runtime errors if a bad UUID is entered in the URL.

frontend/src/utilities.js
mshriver: sometimes I want to ask eslint what it does around here

frontend/src/components/admin-page.js
```diff
@@ -0,0 +1,64 @@
...
+
+
+const AdminPage = (props) => {
+    // TODO useEffect instead of eventEmitter prop
+    // TODO notifications on admin page with state and AlertGroup
+    const navigation = (
+      // TODO what is onNavSelect doing here ... I just carried this from a class ref
+      <PageSidebar theme="dark" >
+        <PageSidebarBody>
+          <Nav onSelect={React.Component.onNavSelect} theme="dark" aria-label="Nav">
```
jjaquish, line 26: It seems that any of the admin or profile pages do not react to changes in dark mode, though changing it there does carry the changes over when switching to a page that does react to it

frontend/src/components/ibutsu-header.js
jjaquish, line 1: I noticed the about page style is a little messed up now. Not sure exactly why

jjaquish (commented):
Left a few comments for now. Functionally it looks fine to me so far (I have some more testing I'd like to do on Monday though) but there were a few style issues

frontend/src/components/admin-page.js
mshriver: does the behavior differ from prod ibutsu right now?

mshriver: Just checked - I did in fact break this here, will take a look at how to get it working better.

mshriver: https://www.patternfly.org/developer-resources/dark-theme-handbook#best-practices

Going to go over the entire app's handling of dark mode following these practices.

mshriver: Pushed a commit to address the theme handling. It might still be a little buggy though with re-renders and syncing to browser default on first render.

frontend/src/components/ibutsu-header.js
mshriver: Can you provide a bit more context? screenshots against production ibutsu on 2.5.8?

frontend/src/components/fileupload.js
```diff
@@ -63,14 +71,14 @@ export class FileUpload extends React.Component {
...
   render() {
     const { children, className } = this.props;
-    const Component = this.props.component || 'button';
     const styles = this.props.isUnstyled ? {} : {cursor: 'pointer', display: 'inline', padding: '0', margin: '0'};
+    const { primaryObject } = this.context;
     return (
       <React.Fragment>
         <input type="file" multiple={this.state.multiple} style={{display: 'none'}} onChange={this.onFileChange} ref={this.inputRef} />
-        <Component className={className} style={styles} onClick={this.onUploadClick}>
+        <Button className={className} style={styles} onClick={this.onUploadClick} isDisabled={!primaryObject}>
```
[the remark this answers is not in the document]
mshriver: I changed this to use `isAriaDisabled` so that it would remain focus-able for accessibility.

The button has been unstyled in ibutsu, and because it's in the header with a plain variant specified, I would agree it's not easy to see that it is disabled until you mouse over it.

Going to commit what I have with the tooltip to provide context, let me know if the styling when disabled is a blocking issue for you and we can discuss it further.

mshriver:
@LightOfHeaven1994 want to take another look with the latest commits?

LightOfHeaven1994:
@mshriver will do tomorrow :+1:

mshriver:
I'm going to proceed with merging and squashing commits, thanks for the reviews and input!

I'll be testing the master branch DB migration against an internal RH stage instance before tagging a release with this commit.

PR #515, opened 2024-10-18
Fix dashboard clearing on project change

mshriver:
The dashboard widgets weren't clearing on project change because the context still held the dashboard from the previous project.

### DIFF
--- a/frontend/src/app.js
+++ b/frontend/src/app.js
@@ -3,7 +3,7 @@ import React from 'react';
 import EventEmitter from 'wolfy87-eventemitter';
 import ElementWrapper from './components/elementWrapper';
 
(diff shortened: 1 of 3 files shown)


frontend/src/app.js
```diff
@@ -85,8 +85,8 @@ export class App extends React.Component {
             path="view/:view_id"
             element={<ElementWrapper routeElement={View} />}
           />
+        <Route path="*" element={<Navigate to="dashboard" replace />} />
```
mshriver, line 88: This takes bare URLs in the tree and redirects to dashboard.

This means if you go to a project with no path afterward, you'll automatically get the dashboard page.

Issue #517, opened 2024-10-23
Migrate to Flask-SQLAlchemy 3


mshriver:
https://flask-sqlalchemy.palletsprojects.com/en/3.1.x/changes/#version-3-0-0

Issue #516, opened 2024-10-23
Migrate to connexion3


mshriver:
https://connexion.readthedocs.io/en/stable/v3.html#migrating-from-connexion-2

PR #518, opened 2024-10-23
[IQE-3104] Update packages

mshriver:
Updating dependencies to address clair scan results from quay.io

### Current Image
https://quay.io/repository/ibutsu/backend/tag/main?tab=securityreport

### PR Image
https://quay.io/repository/ibutsu/backend/tag/pr-518?tab=securityreport

- [x] move dependency pins using uv
- [x] containerfile update
- [x] image builds and local/stage testing

https://issues.redhat.com/browse/IQE-3140


LightOfHeaven1994 (approved):
@mshriver generally looks good, thanks for putting doc lines for the workflow! But we need to make sure different package versions are not breaking our app. I remember that some versions were pinned on purpose

mshriver:
I left connexion, flask, and sqlalchemy pinned where they currently are because the updates for these packages are major, and will break things.

I opened #516  and #517 to capture the need to migrate our implementation to update to the next version.

Everything else I went over the changelog for the project and verified there weren't (documented) breaking changes.

I'm going to deploy these images to stage and test there for behavioral changes or errors popping up.

mshriver:
I did include a move here from `python-jose` to pyJWT, as we were only using the encode/decode functions which are provided by the more actively supported project.

PR #526, opened 2024-10-31
Use source instead of build number when filtering results in Importance by Component widget

Fynardo:
Importance by Component widget uses the _job name_ to fetch runs and the _build number_ and _component_ to group the results. However, there are cases where different jobs may have the same build number and component (for example, two different jobs for stage and prod that contains results for the same component).

This PR modifies the field used in the query from `build_number` to `source` which has more information to not mix runs.

### DIFF
--- a/backend/ibutsu_server/widgets/importance_component.py
+++ b/backend/ibutsu_server/widgets/importance_component.py
@@ -25,16 +25,14 @@ def get_importance_component(
     ).get("jobs")
 
     # A list of job build numbers to filter our runs by
-    job_ids = []
-    for job in jobs:
-        job_ids.append(job["build_number"])
+    job_ids = [job["source"] for job in jobs]
 
     # query for RUN ids
     # metadata has to have a string to column to work
     # because it is a sqlalchemy property otherwise (AFAIK)
     bnumdat = string_to_column("metadata.jenkins.build_number", Run)
     run_data = (
-        Run.query.filter(bnumdat.in_(job_ids), Run.component.in_(components.split(",")))
+        Run.query.filter(Run.source.in_(job_ids), Run.component.in_(components.split(",")))
         .add_columns(Run.id, bnumdat.label("build_number"))
         .all()
     )
--- a/backend/ibutsu_server/widgets/importance_component.py
+++ b/backend/ibutsu_server/widgets/importance_component.py
@@ -1,10 +1,10 @@
-from sqlalchemy import desc, func
 from collections import defaultdict
+
+from sqlalchemy import desc
 
 from ibutsu_server.db.base import session
 from ibutsu_server.db.models import Result, Run
 from ibutsu_server.filters import string_to_column
-from ibutsu_server.widgets.jenkins_job_view import get_jenkins_job_view
 
 
 def get_importance_component(
--- a/backend/ibutsu_server/widgets/importance_component.py
+++ b/backend/ibutsu_server/widgets/importance_component.py
@@ -1,6 +1,7 @@
+from sqlalchemy import desc, func
 from collections import defaultdict
 
-from ibutsu_server.constants import BARCHART_MAX_BUILDS, JJV_RUN_LIMIT
+from ibutsu_server.db.base import session
 from ibutsu_server.db.models import Result, Run
 from ibutsu_server.filters import string_to_column
 from ibutsu_server.widgets.jenkins_job_view import get_jenkins_job_view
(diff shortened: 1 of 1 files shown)


mshriver (approved):

mshriver:
@Fynardo overall I think it's important for us to move away from `jenkins` specific differentiation in the widgets, and support filtering of test results based on abstractions that allow for users to leverage Jenkins job metadata.

mshriver (changes requested):
The widget's configuration is a job name, and the `jenkins_job_view` is providing the job objects here. I would not change this to use `source` without also changing the widget to accurately reflect what is filtering the runs - it's no longer jenkins metadata at that point.

The fact that insights jenkins jobs set a source field that includes the job name is not something that can be applied or assumed.

image

If you want to continue the association with the actual job name for the widget's frontend configuration, I would recommend using the `metadata.jenkins.job_name` field for run filtering, so that the run filtering is explicit.

In fact, `get_jenkins_job_view` will no longer be required here, since the `job_name` is a kwarg.  Given the inline comments, there may be a sqlalchemy concern here - give it a shot!

Fynardo:
@mshriver If we forget for a moment about the fact that I just destroyed the branch (that is why you don't update things at 9pm), I simplified the SQL code that finds and filters the appropriate Jenkins runs. 

I'll try to fix this tomorrow

Fynardo:
@mshriver The code that creates the result dict could be a bit prettier but:

- The complexity is fine, it runs through the result list just once, so at least it is linear.
- I think we would like to address result objects construction using some sort of DTO definitions and make it more formal. However, that would require more design than the one scoped here. We could discuss it further in follow-up PRs

backend/ibutsu_server/widgets/importance_component.py
```diff
@@ -127,7 +115,7 @@ def get_importance_component(
                     }
                 else:
                     sdatdict[component][bnum][importance] = {
-                        "percentage": 0,
+                        "percentage": "-",
```
mshriver, line 118: Consider using `"N/A"`

mshriver:
image

Local testing with compliance job data from an API job and UI job with the same build number and component is looking good, correct calculation and filtering of results from the relevant runs for the filtered job.

mshriver (approved):

mshriver (changes requested):
We need to discuss the SQL changes here, the query is no longer filtering by project since `get_jenkins_job_view` was removed.

backend/ibutsu_server/widgets/importance_component.py
```diff
@@ -1,64 +1,58 @@
...
     )
 
-    # get a list of the job IDs
-    run_info = {}
-    for run in run_data:
-        run_info[run.id] = run.build_number
-
+    # Filter the results based on the jenkins job, build number and component
     mdat = string_to_column("metadata.importance", Result)
+    bnumdat = string_to_column("metadata.jenkins.build_number", Result)
```
mshriver, line 25: These variable names are consistent with other modules, but are needlessly abbreviated.

No need to change them here, this is subjective. But if they bother you as much as they've bothered me, might as well make all of this read easier.

`build_num` and `job_name` would be fine variable names, maybe add a `[]_column`. All they're used for is to reference the label later on, so you could just do that here anyway and chain it.

`build_number = string_to_column("metadata.jenkins.build_number", Result).label("build_number")`

`.add_columns(<>, <>, build_number)`

backend/ibutsu_server/widgets/importance_component.py
```diff
@@ -1,64 +1,58 @@
...
+    jnamedat = string_to_column("metadata.jenkins.job_name", Result)
```
mshriver, line 26: this is no longer used, because the job name is in the run subquery

mshriver: Reached out in slack about this, because if the metadata is on the result, why do we need to subquery the run at all?

What this widget cares about is aggregating by jenkins job and build number - the run is rather irrelevant.

backend/ibutsu_server/widgets/importance_component.py
Fynardo: done

backend/ibutsu_server/widgets/importance_component.py
Fynardo: Removed all references to Runs table in favor of only querying the Results table, using metadata present in the Results table to fetch build_number, job_name, project and component

mshriver (approved):
I'd like to verify this in stage with a large amount of data.

Local testing is looking good!

PR #532, opened 2024-11-19
Update yarn.lock

mshriver:
could be some landmines in here


mshriver:
Looking fine in local testing on build and run.

PR #533, opened 2024-11-19
Fix bug with dashboard selection on back nav

mshriver:
The state should not be updated by the get_default_dashboard function


frontend/src/dashboard.js
```diff
@@ -164,11 +164,6 @@ export class Dashboard extends React.Component {
         'selectedDashboard': targetObject.defaultDashboard,
         'dashboardInputValue': targetObject.defaultDashboard?.title
       })
-    } else {
```
mshriver, line 167: This is only one part of it - there are other dashboard selection bugs with widget fetching of the default dashboard

PR #534, opened 2024-12-03
Convert dashboard to functional

mshriver:
Conversion of the Dashboard component to React functional.

1. Allows for direct use of React and router-dom hooks.
2. Better behavior for utilizing the URL params, simpler filtering behavior.
3. Many 1-line functions moved to callbacks within the JSX.
4. useEffect replacing componentDidMount, with direct correlation to the relevant state or contextual dependency.
5. Removal of `ActiveDashboard` from the context, now correctly implemented with state in Dashboard
6. Addition of `defaultDashboard` to the context, now correctly set when project selection changes

https://issues.redhat.com/browse/IQE-3256

### DIFF
--- a/frontend/src/components/ibutsu-header.js
+++ b/frontend/src/components/ibutsu-header.js
@@ -217,7 +217,12 @@ export class IbutsuHeader extends React.Component {
   }
 
   onProjectSelect = (_event, value) => {
-    const { primaryObject, setPrimaryObject, setPrimaryType } = this.context;
+    const {
+      primaryObject,
+      setPrimaryObject,
+      setPrimaryType,
+      setDefaultDashboard,
+    } = this.context;
     if (primaryObject?.id === value?.id) {
       this.setState({
         isProjectSelectorOpen: false,
@@ -227,8 +232,10 @@ export class IbutsuHeader extends React.Component {
       return;
     }
     // update context
-    setPrimaryObject(value)
-    setPrimaryType('project')
+    setPrimaryObject(value);
+    setPrimaryType('project');
+    setDefaultDashboard(value.default_dashboard_id);
+
     // update state
     this.setState({
       selectedProject: value,
--- a/frontend/src/dashboard.js
+++ b/frontend/src/dashboard.js
@@ -155,7 +155,7 @@ function Dashboard() {
     // state update
     setSelectedDB();
     setIsDBSelectorOpen(false);
-    setSelectDBInputValue('Select a dashboard');
+    setSelectDBInputValue('');
     setFilterDBValue('');
 
     navigate('/project/' + project_id + '/dashboard/')
@@ -218,10 +218,8 @@ function Dashboard() {
     }
     HttpClient.put([Settings.serverUrl, 'widget-config', currentWidget], '', editedData)
         .then(response => HttpClient.handleResponse(response))
-        .then(() => {
-          setIsEditModalOpen(false);
-        })
         .catch(error => console.log(error));
+    setIsEditModalOpen(false);
   }
 
   function onEditWidgetClick(id) {
--- a/frontend/src/dashboard.js
+++ b/frontend/src/dashboard.js
@@ -74,6 +74,7 @@ function Dashboard() {
   const [filterDBValue, setFilterDBValue] = useState('');
 
   useEffect(() => {
+    // update widgets
     let api_params = {'type': 'widget'};
     if (selectedDB) {
       api_params['filter'] = 'dashboard_id=' + selectedDB.id;
@@ -123,7 +124,7 @@ function Dashboard() {
     }
 
 
-  }, [filterDBValue, isDeleteDBOpen, defaultDashboard, primaryObject, navigate, selectedDB]);
+  }, [filterDBValue, defaultDashboard, primaryObject, navigate, selectedDB]);
 
   useEffect(() => {
     // sync the URL to the dashboard selection
@@ -171,22 +172,18 @@ function Dashboard() {
       .then(response => HttpClient.handleResponse(response))
       .then(data => {
         setIsNewDBOpen(false);
-        setSelectedDB(data);
-        setSelectDBInputValue(data.title);
+        onDashboardSelect(null, data);
       })
       .catch(error => console.log(error));
   }
 
   function onDeleteDashboard() {
-
     HttpClient.delete([Settings.serverUrl, 'dashboard', selectedDB.id])
         .then(response => HttpClient.handleResponse(response))
-        .then(() => {
-          setIsDeleteDBOpen(false);
-          setSelectedDB(null);
-          setSelectDBInputValue();
-        })
+        .then(() => onDashboardClear()) // only run after successful delete, otherwise it's a race to remove from the dropdown
         .catch(error => console.log(error));
+    setIsDeleteDBOpen(false);
+
   }
 
   function onDeleteWidgetClick(id) {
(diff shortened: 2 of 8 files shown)


frontend/src/components/edit-widget-modal.js
```diff
@@ -153,7 +153,7 @@ export class EditWidgetModal extends React.Component {
               </HelperText>
             </FormHelperText>
           </FormGroup>
-          {componentLoaded ? widgetType.params.map(param => {
+          {componentLoaded ? widgetType?.params.map(param => {
```
Andrewgdewar, line 156: if widgetType doesn't exist, you will return the string "undefined" in the dom.

I'd suggest (widgetType?.params||[]).map

Or 

componentLoaded && !!widgetType ?

frontend/src/components/ibutsu-header.js
```diff
@@ -227,8 +232,11 @@ export class IbutsuHeader extends React.Component {
       return;
     }
     // update context
-    setPrimaryObject(value)
-    setPrimaryType('project')
+    setPrimaryObject(value);
+    setPrimaryType('project');
+    console.log('project selected: ' + value.default_dashboard_id)
```
Andrewgdewar, line 237: rogue console.

frontend/src/dashboard.js
```diff
@@ -46,623 +44,528 @@ import {
...
 
-export class Dashboard extends React.Component {
-  static contextType = IbutsuContext;
-  static propTypes = {
-    eventEmitter: PropTypes.object,
-    navigate: PropTypes.func,
-    params: PropTypes.object,
-  }
+function Dashboard() {
+  const context = useContext(IbutsuContext);
```
Andrewgdewar, line 51: I'd suggest just destructuring what you need here directly (#nit)

frontend/src/dashboard.js
```diff
@@ -46,623 +44,528 @@ import {
...
+  // typeahead input value states
+  const [selectDBInputValue, setSelectDBInputValue] = useState('');
+  const [filterDBValue, setFilterDBValue] = useState('');
+
+  useEffect(() => {
+    // fetch widgets with the selected dashboard changes
+
+    // TODO is sync necessary with functional component...?
+    //sync_context();
+    console.log('dashboard useEffect selectedDB');
```
Andrewgdewar, line 81: Guessing you want to leave these then?

frontend/src/dashboard.js
```diff
@@ -46,623 +44,528 @@ import {
...
-    if ( paramDash != null && activeDashboard?.id !== paramDash) {
-      HttpClient.get([Settings.serverUrl, 'dashboard', paramDash])
+  }, [defaultDashboard])
+
+  const { dashboard_id } = params;
+  useEffect(() => {
+    // sync the URL to the dashboard selection
+    if (selectedDB?.id !== dashboard_id) {
+      console.log('dashboard useEffect dashboard_id');
+      HttpClient.get([Settings.serverUrl, 'dashboard', dashboard_id])
```
Andrewgdewar, line 117: Not so much something to fix, but just a pattern you can use. 
Self evoked functions allow for async usage ie you could do:

(async()=>{
try {
const result = await HttpClient.get([Settings.serverUrl, 'dashboard', dashboard_id])
// do things
} catch (e){
handle catch
}


} )()

frontend/src/dashboard.js
```diff
@@ -46,623 +44,528 @@ import {
...
+              <TimesIcon aria-hidden />
+            </Button>
+          )}
+        </TextInputGroupUtilities>
+      </TextInputGroup>
+    </MenuToggle>
+  )
+
+  return (
+    <React.Fragment>
```
Andrewgdewar, line 315: can be simiplied to '<>' and '</>'

Andrewgdewar:
I would suggest getting eslint and prettier installed on your vs-code for those tools to help you simplify and catch a lot of small nit/spacing/simplifications.

frontend/src/components/edit-widget-modal.js
mshriver: thanks!

frontend/src/components/ibutsu-header.js
mshriver: thx, forgot to drop all these before marking ready.

frontend/src/dashboard.js
mshriver: nah removing, they were for my lifecycle observation while working on it

frontend/src/dashboard.js
mshriver: absolutely makes sense, I started with this but don't really use it "directly".

mshriver: Looks like I can apply the same pattern to my use of `params` hook as well.

frontend/src/dashboard.js
mshriver: Might have some further questions about this one, but it makes sense. I think I'm actually 'suffering' from this in #535, when the context updates and triggers the request, the rest of the useEffects run before the response, and then the updated state from the response doesn't trigger a render, so the state is updated but the rendered view is out of sync.

mshriver:
Thanks for the review @Andrewgdewar ! I've applied most of the suggestions, and am going to look at the self-calling await for my fetches.  I'm also interested in moving those HTTP calls to the react-router-dom loader hooks instead, and maybe would use the same await pattern there?

eslint is part of the pre-commit config for the repo, maybe it needs some config updates?

mshriver:
I opened #536 to update eslint, and add some plugins.

frontend/src/dashboard.js
```diff
@@ -46,619 +44,493 @@ import {
...
+      editedData.project_id = primaryObject.id;
     }
-    this.setState({isEditModalOpen: false});
-    editWidget.id = this.state.currentWidgetId
-    HttpClient.put([Settings.serverUrl, 'widget-config', this.state.currentWidgetId], '', editWidget)
+    HttpClient.put([Settings.serverUrl, 'widget-config', currentWidget], '', editedData)
         .then(response => HttpClient.handleResponse(response))
         .then(() => {
-          this.getWidgets();
+          setIsEditModalOpen(false);
```
LightOfHeaven1994, line 222: @mshriver we should move `setIsEditModalOpen(false);` out from HTTP request as it's made async and as the result our modal is still marked as open for some time and it cases modal re-render issue.
How to reproduce:

1. Open edit widget modal
2. Try to change any value
3. Click save
As the result you get an error as
```
  onSave = () => {
    const updatedWidget = {
      title: this.state.title,
      params: this.state.params,
      weight: parseInt(this.state.weight),
      type: 'widget',
      widget: this.props.data.widget
    }
    this.props.onSave(updatedWidget);
    console.log("DEBUG 2")
    this.setState({
      widgetType: null,
      title: '',
      params: {},
      weight: 0,
      isTitleValid: false,
      areParamsFilled: false
    });
  }
  ``` in edit-widget-modal.js clears the state but modal is still set as open.
  
  To fix this issue:
  ```
    function onEditWidgetSave(editedData) {
      if (!editedData.project_id && primaryObject) {
        editedData.project_id = primaryObject.id;
      }
      HttpClient.put([Settings.serverUrl, 'widget-config', currentWidget], '', editedData)
          .catch(error => console.log(error));
      setIsEditModalOpen(false);
    }
  ```

LightOfHeaven1994 (commented):
there are 2 things I found, other than that - looks good to me, great work!

frontend/src/dashboard.js
mshriver: great catch - I'm going to keep an eye for these cases where there's no dependence on the returned data.

mshriver: Updated onDeleteDashboard too, and actually moved the state calls to before the promise.

mshriver:
As this is going into a feature branch and not main, I'm going to merge with the existing review feedback.

This will continue to be tested in the feature branch with #535, and the branch will be deployed to stage for scale testing.

PR #535, opened 2024-12-10
Convert IbutsuHeader to functional component

mshriver:
Introduce react-toastify in addition to the migration, in order to get rid of the eventEmitter and improve the notifications pattern.

The ToastContainer has been placed in the IbutsuPage and ProfilePage, where the notifications are relevant.

Notification hooks have been moved from the header to the actual FileUpload that produces them. I left FileUpload as a class component but removed the beforeUpload and afterUpload hooks as I don't believe it makes sense to inject notification hooks through those function callbacks.

- [x] Theme toggle is too sticky for browser/OS preference, stays dark
- [x] yarn test failure, `    TypeError: window.matchMedia is not a function`, is occurring on an unmodified line of code in the utility function. It works locally.

### DIFF
--- a/frontend/src/components/FileUpload.spec.cy.js
+++ b/frontend/src/components/FileUpload.spec.cy.js
@@ -2,7 +2,6 @@ import { mount } from 'cypress/react';
 import React from 'react';
 
 import { FileUpload } from '.';
-import { IbutsuContext } from '../services/context';
 
 describe('FileUpload', () => {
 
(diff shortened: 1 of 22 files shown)


frontend/src/pages/profile/user.js
```diff
@@ -87,12 +80,12 @@ export class UserProfile extends React.Component {
     let tempUser = Object.assign({}, user, {name: tempName});
     this.saveUser(tempUser).then(response => {
       if (response !== undefined) {
-        this.showNotification('success', 'Name Updated', 'Your name has been updated.');
+        // this.showNotification('success', 'Name Updated', 'Your name has been updated.');
```
Fynardo, line 83: We may want to remove this line entirely instead of commenting? Maybe launch a new toast?

mshriver: Let's call it early review state :smile: 

This will get replaced with `toast.success()` in this PR, but I'd love some feedback on the integration, styling, and patterns before using it here too.

frontend/src/components/FileUpload.spec.cy.js
```diff
@@ -11,35 +10,4 @@ describe('FileUpload', () => {
     cy.get('button');
   });
 
-  it('should fire the beforeUpload when a file is changed', () => {
```
mshriver, line 14: this test was providing very little value as-is, but the hooks it's checking have been removed.

The testing coverage for the app needs to be addressed holistically.

PR #536, opened 2024-12-12
Node 20, ESLint 9

mshriver:
Updating to eslint9, including prettier (no config yet), and moving to new config format for eslint.

- [x] cypress/react18 warning. 
- [x] *Investigate the file upload test failure.* Upload works in the running app right now, it's just the test.
- [x] Frontend build failure


frontend/src/run.js
```diff
@@ -752,16 +747,10 @@ export class Run extends React.Component {
                   </CardBody>
                 </Card>
               </Tab>
-              <Tab eventKey={'results-tree'} title={<TabTitle icon={RepositoryIcon} text="Results Tree" />}>
+              <Tab eventKey="results-tree" title={<TabTitle icon={RepositoryIcon} text="Results Tree" />}>
                 <Card className="pf-u-mt-lg">
                   <CardBody>
                     <Grid gutter="sm">
-                      {false && <GridItem span={12}>
```
mshriver, line 759: @jjaquish do you know what you were doing here with `false &&` syntax?

frontend/src/views/accessibilityanalysis.js
```diff
@@ -465,44 +463,15 @@ export class AccessibilityAnalysisView extends React.Component {
                     width={300}
                   />
               </div>
-
-              }
-              {false &&
```
mshriver, line 470: @jjaquish same as above, why the `bool && <element>` here - in this case with a true and a false?

Andrewgdewar: Eliminate it!

PR #546, opened 2025-01-23
Convert ClassifyFailureTable to functional

mshriver:
- [x] Row expansion isn't working right now
- [x] yarn lint with unused states/setState hooks
- [x] MetaFilter errors on setFilter
- [x] Bug in main branch, error on expanding row: 
image

Working multi-classification and single classification

Tested on run classify failures tab, and on result page.

### DIFF
--- a/frontend/src/components/classify-failures.js
+++ b/frontend/src/components/classify-failures.js
@@ -1,8 +1,7 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 
 import {
-  Button,
   Card,
   CardHeader,
   CardBody,
@@ -10,7 +9,8 @@ import {
   Flex,
   FlexItem,
   TextContent,
-  Text,
+  Title,
+  Badge,
 } from '@patternfly/react-core';
 import {
   TableVariant,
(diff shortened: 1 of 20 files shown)


frontend/src/components/classify-failures.js
```diff
@@ -80,184 +151,173 @@ export class ClassifyFailuresTable extends React.Component {
...
+          };
+        } else {
+          newRow = {...row};
+        }
+        return newRow;
       });
+      setRows(updatedRows);
+    } else {
+      // handle updating isOpen on the clicked row (closing)
+      setRows(rows.map((row, index) => {
```
LightOfHeaven1994, line 183: instead of going through each row, we can find row that has been opened and set isOpen to true and ignore other rows:
```
      setRows(prevRows => {
        const updatedRows = [...prevRows];
        if (updatedRows[rowIndex]) {
          updatedRows[rowIndex] = { ...updatedRows[rowIndex], isOpen };
        }
        return updatedRows;
      });
```

frontend/src/components/classify-failures.js
```diff
@@ -20,56 +20,127 @@ import {
...
+  useEffect(()=> {
+    setIsError(false);
 
-  onCollapse (event, rowIndex, isOpen) {
-    const { rows } = this.state;
+    // get only filtered results
+    let params = {};
+    params['filter'] = toAPIFilter(appliedFilters);
+    params['pageSize'] = pageSize;
+    params['page'] = page;
```
LightOfHeaven1994, line 124: nitpick
```
  const params = {
    filter: toAPIFilter(appliedFilters),
    pageSize,
    page,
  };
  ```

File: frontend/src/components/classify-failures.js [hunk 3]
```diff
@@ -80,184 +151,173 @@ export class ClassifyFailuresTable extends React.Component {
...
-        this.setState({rows: [], isEmpty: false, isError: true});
-      });
-  }
+  useEffect(() => {
+    // set rows when filtered items update
+    setRows(
+      filteredResults.map(
+        (result, index) => resultToClassificationRow(result, index, setFilter)
+      )
+        .flat()
```
LightOfHeaven1994, line 270: instead of doing map() and then flat() - you can do flatMap()
```
setRows(
  filteredResults.flatMap((result, index) => resultToClassificationRow(result, index, setFilter))
);
```

frontend/src/components/metafilter.js
```diff
@@ -0,0 +1,222 @@
...
+    onRemoveFilter,
+    onApplyReport,
+    runId,
+    id
+  } = props;
+  const context = useContext(IbutsuContext);
+  const {primaryObject} = context;
+
+  const [fieldSelection, setFieldSelection] = useState([]);
+  const [isFieldOpen, setisFieldOpen] = useState(false);
```
LightOfHeaven1994, line 41: nitpick: should be `setIsFieldOpen`

frontend/src/components/metafilter.js
```diff
@@ -0,0 +1,222 @@
...
+  }, [primaryObject.id]);
+
+  let values_available = valueOptions.length > 0;
+  let value_placeholder = 'Select a field first' ; // default instead of an else block
+  if (fieldSelection.length > 0 && values_available) {
+    value_placeholder = 'Select value(s)';
+  }
+  else if (fieldSelection.length > 0 && !values_available) {
+    value_placeholder = 'No values for selected field';
+  }
```
LightOfHeaven1994, line 135: nitpick we can replace it with:
```
  const valuePlaceholder = !fieldSelection.length
  ? 'Select a field first'
  : values_available
    ? 'Select value(s)'
    : 'No values for selected field';
```
also vars can be camelCase

frontend/src/components/metafilter.js
```diff
@@ -0,0 +1,222 @@
...
+      : [...valueSelections, selection];
+
+    setValueSelections(updatedValues);
+    setIsValueOpen(false);
+    if (updatedValues.length > 0) {
+      setFilter(id, fieldSelection, updatedValues.join(';'));
+    }
+
+    // after filter was applied, remove the values from the fields
+    setValueSelections([]);
```
LightOfHeaven1994, line 71: tbh it looks a bit weird that we setValueSelection couple rows above and then clearing it. I was following behavior and I got that after you select filter and value - it's set and chip appears but select clears itself (that's fine) but we maybe don't need all of this set thing here?

LightOfHeaven1994 (approved):
generally looks good and everything is working. But I left some suggestions

frontend/src/components/metafilter.js
mshriver: Without setting it back to an empty list the value chip stays in the dropdown (even though it's been added to the applied filters) - maybe I'm missing your suggestion for where you'd prefer to see it?

LightOfHeaven1994: @mshriver maybe we don't need to have `setValueSelections(updatedValues);` couple lines above

frontend/src/components/metafilter.js
mshriver: heh, I looked at this block for a moment and thought about golfing it but decided to ignore it :)

frontend/src/components/metafilter.js
mshriver: Talked more in slack - @LightOfHeaven1994 is correct that we don't actually need to track or set the `selections` prop because on value selection the filter is applied and the select is cleared anyway.

PR #551, opened 2025-02-04
Convert EditWidgetModal

mshriver:
## TODO
- [x] ~Investigate empty title/weight on first render of modal, subsequent views render title first time~
- [x] ~Accessibility widget has params, adding one fails, need to fix that before I can test validation of required parameter fields~

### DIFF
--- a/frontend/src/components/edit-widget-modal.js
+++ b/frontend/src/components/edit-widget-modal.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 
 import {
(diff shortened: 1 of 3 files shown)


frontend/src/components/edit-widget-modal.js
```diff
@@ -20,168 +20,156 @@ import { Settings } from '../settings';
...
+          <TextInput type="number" id="widget-weight" name="widget-weight" value={weight} onChange={(_event, value) => setWeight(value)} />
+          <FormHelperText>
+            <HelperText>
+              <HelperTextItem variant="default">
+              How widgets are ordered on the dashboard
+              </HelperTextItem>
+            </HelperText>
+          </FormHelperText>
+        </FormGroup>
+        {componentLoaded ? widgetType.params.map(param => (
```
mshriver, line 134: this isn't working right, no params getting rendered right now.

PR #559, opened 2025-02-12
Convert TableStates

mshriver:


frontend/src/components/filtertable.js
```diff
@@ -26,11 +26,11 @@ import { Settings } from '../settings';
 import { HttpClient } from '../services/http';
 import { toAPIFilter } from '../utilities';
 
-import { TableEmptyState, TableErrorState } from './tablestates';
+import {TableEmptyState, TableErrorState} from './tablestates'
```
LightOfHeaven1994, line 29: @mshriver we should probably have some linter that will autofix all of these spaces and `;`

frontend/src/components/tablestates.js
```diff
@@ -15,58 +15,57 @@ import {
...
-      </Bullseye>
-    );
-  }
+const TableEmptyState = (props) => {
+  const {onClearFilters} = props;
+  <Bullseye>
+    <EmptyState>
+      <EmptyStateHeader titleText="No results found" icon={<EmptyStateIcon icon={SearchIcon} />} headingLevel="h5" />
+      <EmptyStateFooter>
+        {!!this.props.onClearFilters &&
```
LightOfHeaven1994, line 26: it's still using `this.`

frontend/src/components/tablestates.js
```diff
@@ -15,58 +15,57 @@ import {
...
-          </EmptyStateFooter>
-        </EmptyState>
-      </Bullseye>
-    );
-  }
+  <Bullseye>
+    <EmptyState>
+      <EmptyStateHeader titleText="Error occurred fetching results" icon={<EmptyStateIcon icon={ErrorCircleOIcon} />} headingLevel="h5" />
+      <EmptyStateFooter>
+        {!!this.props.onClearFilters &&
```
LightOfHeaven1994, line 52: it's still using `this.`

frontend/src/components/tablestates.js
```diff
@@ -15,58 +15,57 @@ import {
...
+            An error occurred while fetching results. Try a different set of filters.
+          </EmptyStateBody>
+          <EmptyStateActions>
+            <Button variant="link" onClick={onClearFilters}>Clear all filters</Button>
+          </EmptyStateActions>
+        </React.Fragment>
+        }
+      </EmptyStateFooter>
+    </EmptyState>
+  </Bullseye>
```
LightOfHeaven1994, line 64: Both `TableErrorState` & `TableEmptyState` are almost identical, I think we can merge these 2 and add more props to differ them:

```
const TableState = ({ title, icon, message, onClearFilters }) => (
...
);

const TableEmptyState = (props) => (
  <TableState
    title="No results found"
    icon={SearchIcon}
    message="No results match this filter criteria. Clear all filters to show results."
    {...props}
  />
);

const TableErrorState = (props) => (
  <TableState
    title="Error occurred fetching results"
    icon={ErrorCircleOIcon}
    message="An error occurred while fetching results. Try a different set of filters."
    {...props}
  />
);
```

frontend/src/components/tablestates.js
```diff
@@ -15,58 +15,57 @@ import {
...
 
-export class TableErrorState extends React.Component {
-  static propTypes = {
-    onClearFilters: PropTypes.func,
-  }
+TableEmptyState.propTypes = {
+  onClearFilters: PropTypes.func,
+}
+
+const TableErrorState = (props) => {
```
LightOfHeaven1994, line 45: since you are using `{}` we need to have `return`

frontend/src/components/tablestates.js
```diff
@@ -15,58 +15,57 @@ import {
...
-                <Button variant="link" onClick={this.props.onClearFilters}>Clear all filters</Button>
-              </EmptyStateActions>
-            </React.Fragment>
-            }
-          </EmptyStateFooter>
-        </EmptyState>
-      </Bullseye>
-    );
-  }
+const TableEmptyState = (props) => {
```
LightOfHeaven1994, line 20: since you are using `{}` we need to have `return`

LightOfHeaven1994 (commented):
@mshriver we need to fix couple things before merging it, good work 👍

frontend/src/components/filtertable.js
mshriver: correct - I included the prettier dependency in one of the modernization/eslint PRs, but haven't applied it.

Plan is to apply it following this branch merging to main

frontend/src/components/tablestates.js
mshriver: yeah, I put a TODO above the first component mentioning this, but don't want to do it right now. :)

frontend/src/components/tablestates.js
mshriver: ah I wasn't unpacking the props before, added the `const` line and didn't add the return. thanks!

PR #560, opened 2025-02-12
Convert Importance by Component Widget

Fynardo:


frontend/src/widgets/importancecomponent.js
```diff
@@ -24,81 +24,69 @@ import { Settings } from '../settings';
...
           throw Error(response.statusText);
         }
         return response.json();
       })
-      .then(data => this.setState({data: data, isLoading: false}))
+      .then(data => {
+        setTableData(data.table_data);
+        setIsLoading(false);
+        setDataError(false);
+        return data
```
LightOfHeaven1994, line 54: nitpick: No need to return data here, we have already used data to `setTableState`

LightOfHeaven1994 (approved):
generally lgtm, great work!

LightOfHeaven1994 (approved):
generally lgtm, great work!

frontend/src/widgets/importancecomponent.js
Fynardo: fixed, thanks for reviewing!

LightOfHeaven1994 (approved):

PR #567, opened 2025-02-13
Added logic for different keycloak auth paths

wcater803:
In the itless environment, the Keycloak authentication instance was upgraded.  In this upgrade the "/auth" endpoint was deprecated.  This logic keeps the existing logic with the hardcoded "login/auth/keycloak" endpoint, but also adds in the availability to overwrite that by using the keycloakAuthPath (in the frontend) and KEYCLOAK_AUTH_PATH (foir backend) environment variables.

### DIFF
--- a/backend/ibutsu_server/util/keycloak.py
+++ b/backend/ibutsu_server/util/keycloak.py
@@ -26,7 +26,8 @@ def get_keycloak_config(is_private=False):
         "authorization_url": build_url(realm_base_url, "protocol/openid-connect/auth"),
         "realm": realm,
         "client_id": current_app.config.get("KEYCLOAK_CLIENT_ID"),
-        "redirect_uri": backend_url + current_app.config.get("KEYCLOAK_AUTH_PATH", "/login/auth/keycloak")
+        "redirect_uri": backend_url
+        + current_app.config.get("KEYCLOAK_AUTH_PATH", "/login/auth/keycloak"),
     }
     if current_app.config.get("KEYCLOAK_ICON"):
         config["icon"] = current_app.config["KEYCLOAK_ICON"]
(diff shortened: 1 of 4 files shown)


frontend/src/services/keycloak.js
```diff
@@ -6,6 +6,6 @@ export class KeycloakService {
   static login(url, realm, client_id) {
     const keycloakInstance = new Keycloak({url: url, realm: realm, clientId: client_id});
     keycloakInstance.init({onLoad: 'login-required', checkLoginIframe: false, responseMode: 'query',
-                           redirectUri: Settings.serverUrl + '/login/auth/keycloak'});
```
mshriver, line 9: Since the server URL is the same both times, you can flip this logic a bit.

`redirectUri: Settings.serverUrl + Settings.keycloakAuthPath ? Settings.keycloakAuthPath : '/login/auth/keycloak'`

You may be able to use `||` to be even more direct, but I'd have to test it when the envvar is empty.

wcater803: I modified the code to use the or operator, as long as the keycloakAuthPath is a [Falsey](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) value it works fine

backend/ibutsu_server/util/keycloak.py
```diff
@@ -32,6 +32,8 @@ def get_keycloak_config(is_private=False):
         config["icon"] = current_app.config["KEYCLOAK_ICON"]
     if current_app.config.get("KEYCLOAK_NAME"):
         config["display_name"] = current_app.config["KEYCLOAK_NAME"]
+    if current_app.config.get("KEYCLOAK_AUTH_PATH"):
```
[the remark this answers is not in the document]
wcater803: Modified to use more direct logic

mshriver (approved):

PR #568, opened 2025-02-14
Convert projectlist

mshriver:
document is still re-rendering continuously it's just not visible other than the doc title

Table is rendering, filtering is working, delete is working, and there's a new empty state displayed instead of the table 'no results found' view.

### DIFF
--- a/frontend/src/pages/admin/project-list.js
+++ b/frontend/src/pages/admin/project-list.js
@@ -1,5 +1,4 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { useEffect, useState } from 'react';
 
 import {
   Button,
@@ -19,13 +18,30 @@ import { Link } from 'react-router-dom';
 
 import { HttpClient } from '../../services/http';
 import { Settings } from '../../settings';
-import { debounce, getSpinnerRow } from '../../utilities';
+import { getSpinnerRow } from '../../utilities';
 import { FilterTable } from '../../components/filtertable';
+import { EmptyObject } from '../../components';
 
+const COLUMNS = ['Title', 'Name', 'Owner', ''];
 
-function projectToRow (project, onDeleteClick) {
-  return {
-    cells: [
+const ProjectList = () => {
+  const [filterText, setFilterText] = useState('');
+  const [filteredProjects, setFilteredProjects] = useState([]);
+  const [activeFilters, setActiveFilters] = useState({});
+  const [page, setPage] = useState(1);
+  const [pageSize, setPageSize] = useState(20);
+
+  const [projects, setProjects] = useState([]);
+
+  const [totalItems, setTotalItems] = useState(0);
+  const [isError, setIsError] = useState(false);
+  const [isEmpty, setIsEmpty] = useState(false);
+  const [selectedProject, setSelectedProject] = useState();
+  const [isDeleting, setIsDeleting] = useState(false);
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
+
+  const projectToRow = (project) => {
+    const shutupEslint = {'cells': [
       {title: project.title},
       {title: project.name},
       {title: project.owner && project.owner.name},
@@ -44,7 +60,10 @@ function projectToRow (project, onDeleteClick) {
             <Button
               variant="danger"
               ouiaId={`admin-projects-delete-${project.id}`}
-              onClick={() => onDeleteClick(project.id)}
+              onClick={() => {
+                setSelectedProject(project);
+                setIsDeleteModalOpen(true);
+              }}
               size="sm"
             >
               <TrashIcon />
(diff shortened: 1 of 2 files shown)


frontend/src/pages/admin/project-list.js
```diff
@@ -44,216 +60,177 @@ function projectToRow (project, onDeleteClick) {
...
+                component={(props) => <Link {...props} to="/admin/projects/new" />}
+              >
+                <PlusCircleIcon /> Add Project
+              </Button>
+            </FlexItem>
+          </Flex>
+        </Flex>
+      </PageSection>
+      <PageSection className="pf-u-pb-0">
+        {!isEmpty &&
```
LightOfHeaven1994, line 168: We should check here if `projects.length > 0` instead of isEmpty. So if there are no projects at all - we will show EmptyObject with `No Projects found` & `Create your first project`. But if length is more than 0 - we have some projects and now we will show filteredProjects. If it's going to be empty list (in case we used filter - isEmpty will take care of showing table empty thing)

isEmpty is used as param for FilterTable and when it's set to true - table shows No results found
image

frontend/src/pages/admin/project-list.js
```diff
@@ -44,216 +60,177 @@ function projectToRow (project, onDeleteClick) {
...
+              }}
+              isEmpty={isEmpty}
+              isError={isError}
+              onSetPage={(_event, value) => {setPage(value);}}
+              onSetPageSize={(_event, value) => {setPageSize(value);}}
+            />
+          </CardBody>
+        </Card>
+        }
+        {isEmpty && !filterText && <EmptyObject headingText='No Projects found' bodyText='Create your first project' returnLink='/admin/projects/new' returnLinkText='Add Project'/>}
```
LightOfHeaven1994, line 199: It's enough here to check if `projects.length === 0` - it means backend has returned 0 projects and we don't care if filter has been set or not since there are no projects at all

frontend/src/pages/admin/project-list.js
```diff
@@ -44,216 +60,177 @@ function projectToRow (project, onDeleteClick) {
...
+        </Flex>
+      </PageSection>
+      <PageSection className="pf-u-pb-0">
+        {!isEmpty &&
+        <Card>
+          <CardBody className="pf-u-p-0">
+            <FilterTable
+              columns={COLUMNS}
+              rows={filteredProjects ? filteredProjects.map((p) => projectToRow(p)) : [getSpinnerRow(4)]}
+              activeFilters={activeFilters}
```
LightOfHeaven1994, line 174: To make it work correctly we also need to provide `onRemoveFilter={onRemoveFilter}` that will handle chip removing logic. Otherwise when you click on delete chip - page crashes.

onRemoveFilter will need to look like

```
  const onRemoveFilter = () => {
    setFilterText('');
    setActiveFilters({});
    setIsEmpty(false);
    setFilteredProjects(projects);
    setTotalItems(projects.length);
  };
  ```

**Review by LightOfHeaven1994 — COMMENTED** [review]
I took a look locally and proposing some changes. LMK if you have questions, we can have a call and I will explain.

P.S I just found out that we stopped using API calls when filtering projects (but we used it before) so my comments are correct only if we want to leave behavior when we use API calls when paginating and displaying results but using JS when project is filtered by title. TBH, I would ignore my comments and do fixes so we start using project filtering with backend request (and we can remove this useEffect)

So what we can do is `HttpClient.get([Settings.serverUrl, 'admin', 'project'], {page: page, pageSize: pageSize})` add here filtering, and trigger this useEffect on `filterText` change. But you will need to deal with `Create your first project` thing that you added and I have an idea how to do that

File: frontend/src/pages/admin/project-list.js [hunk 4]
```diff
@@ -19,13 +18,30 @@ import { Link } from 'react-router-dom';
...
+
+  const [totalItems, setTotalItems] = useState(0);
+  const [isError, setIsError] = useState(false);
+  const [isEmpty, setIsEmpty] = useState(false);
+  const [selectedProject, setSelectedProject] = useState();
+  const [isDeleting, setIsDeleting] = useState(false);
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
+
+  const projectToRow = (project) => {
+    const shutupEslint = {'cells': [
```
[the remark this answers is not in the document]
mshriver: yeah this is kinda silly, but if I try to put the returned object directly following the `=>` it also complains.

frontend/src/pages/admin/project-list.js
```diff
@@ -44,216 +60,177 @@ function projectToRow (project, onDeleteClick) {
...
+      newProjects = projects.filter(p =>
+        String(p.title).toLowerCase().includes(filterText.toLowerCase())
+      );
+
+      if (newProjects.length === 0) {
+        newProjects = [{
+          isDisabled: true,
+          value: {},
+          title: `No results found for "${filterText}"`,
+        }];
```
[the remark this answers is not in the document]
mshriver: image

Screenshot wasn't attached - this is what I see when filtering by a project that doesn't exist.

frontend/src/pages/admin/project-list.js
```diff
@@ -44,216 +60,177 @@ function projectToRow (project, onDeleteClick) {
...
       });
-  }
+  }, [page, pageSize, isDeleting]); // isDeleteing so the projects fetch after delete
 
-  onDeleteClick = (projectId) => {
-    const selectedProject = this.state.projects.find((project) => project.id === projectId);
-    this.setState({selectedProject: selectedProject, isDeleteModalOpen: true});
+  const onFilterChange = (_event, value) => {
+    setFilterText(value);
+    setActiveFilters(value ? {'title': value} : {});
```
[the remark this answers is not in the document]
mshriver: If I do this, then empty state will be shown when the filter returns no projects - which is not correct.

I can move the effect content into `onFilterChange` since `setTextFilter` is only called within it.

This would remove the behavior of re-running the filter application when `projects` changes though, as it's also a dependency of the effect.

frontend/src/pages/admin/project-list.js
mshriver: AH! didn't remove filter in my testing - good catch, adding.

frontend/src/pages/admin/project-list.js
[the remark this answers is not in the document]
mshriver: Going to keep the useEffect for handling changes to `projects` or `textFilter`, and I actually moved `setActiveFilters` into the useEffect.

frontend/src/pages/admin/project-list.js
mshriver: Updated to use FilterTable empty state here, and kept the custom empty state when no projects are found at all.

LightOfHeaven1994 (approved):
LGTM, great work!

PR #571, opened 2025-02-17
Convert UserEdit to functional

mshriver:


frontend/src/pages/admin/user-edit.js
```diff
@@ -19,337 +18,287 @@ import {
...
+          isDisabled: true,
+          value: {},
+          title: `No results found for "${inputValue}"`,
+        }];
       }
 
-      this.setState({
-        filteredProjects: newSelectOptions,
-      });
+      if (!isProjectsOpen){
```
LightOfHeaven1994, line 127: We don't need to have it in useEffect as MenuToggle handles it

frontend/src/pages/admin/user-edit.js
```diff
@@ -19,337 +18,287 @@ import {
...
+        setIsProjectsOpen(true);
+      }
     }
-  }
+    setFilteredProjects(newProjectOptions);
 
-  render () {
-    const { projects, inputValue, filteredProjects, user, userProjects } = this.state;
+  }, [inputValue, projects, isProjectsOpen]);
```
LightOfHeaven1994, line 133: `isProjectsOpen` can be removed from dependencies

LightOfHeaven1994 (approved):
Generally looks good, I left couple suggestions

frontend/src/pages/admin/user-edit.js
mshriver: Only if I remove forcing it open when there is an input value present. ;)

LightOfHeaven1994: @mshriver oh I see, some specifics because of typeahead, then ignore my comments

PR #572, opened 2025-02-18
Convert ResultAggregator widget

Fynardo:


frontend/src/widgets/resultaggregator.js
```diff
@@ -22,176 +22,166 @@ import WidgetHeader from '../components/widget-header';
...
+    'var(--pf-v5-global--success-color--100)',
+    'var(--pf-v5-global--danger-color--100)',
+    'var(--pf-v5-global--warning-color--100)',
+    'var(--pf-v5-global--info-color--100)',
+    'var(--pf-v5-global--primary-color--100)'
+  ];
+
+  return (
+    <Card>
+      <WidgetHeader title={title} getDataFunc={getChartData} onEditClick={onEditClick} onDeleteClick={onDeleteClick}/>
```
mshriver, line 96: If you move `getChartData` to a direct `useEffect` hook, this prop would be removed and `WidgetHeader` just won't render a refresh button.

frontend/src/widgets/resultaggregator.js
```diff
@@ -22,176 +22,166 @@ import WidgetHeader from '../components/widget-header';
...
+      <CardBody data-id="recent-result-data">
+        {resultAggregatorError &&
+          <p>Error fetching data</p>
+        }
+        {(total === 0 && !isLoading) &&
+          <p>No data returned, try changing the days.</p>
+        }
+        {(!resultAggregatorError && isLoading) &&
+        <Text component="h2">Loading ...</Text>
```
mshriver, line 105: Consider `<Title headingLevel='h2'>`

frontend/src/widgets/resultaggregator.js
```diff
@@ -22,176 +22,166 @@ import WidgetHeader from '../components/widget-header';
...
+        />
+        }
+        <ParamDropdown
+          dropdownItems={dropdownItems || ['result', 'metadata.exception_name', 'component', 'metadata.classification']}
+          defaultValue={params.group_field}
+          handleSelect={onGroupFieldSelect}
+          tooltip="Group data by:"
+        />
+        <ParamDropdown
+          dropdownItems={[0.1, 0.5, 1, 3, 5]}
```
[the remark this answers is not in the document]
Fynardo: I went ahead and added some values that could fit as testing

frontend/src/widgets/resultaggregator.js
Fynardo: Done

frontend/src/widgets/resultaggregator.js
Fynardo: Done

frontend/src/widgets/resultaggregator.js
```diff
@@ -22,176 +22,166 @@ import WidgetHeader from '../components/widget-header';
...
+  const [chartData, setChartData] = useState([]);
+  const [legendData, setLegendData] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [total, setTotal] = useState(0);
+  const [resultAggregatorError, setResultAggregatorError] = useState(false);
 
-  getResultData () {
-    this.setState({isLoading: true});
-    HttpClient.get([Settings.serverUrl, 'widget', 'result-aggregator'], this.params)
+  const getChartData = useCallback(() => {
```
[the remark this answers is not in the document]
Fynardo: Best I could get is a dependency array of `params`, `days` and `groupFields`. It is not awesome code though.
Issue is, if we set `params` as only dependency then we either:

1. Modify inner fields (e.g `params.<field> = value`) which won't modify the object, thus, `useEffect` will not run.
2. Create a new `params` object (e.g. `params = {...params, <field>: value}` which will fail because params is read-only

frontend/src/widgets/resultaggregator.js
```diff
@@ -22,176 +22,164 @@ import WidgetHeader from '../components/widget-header';
...
-      isLoading: true,
-      total: 0,
-    };
-  }
+  const [chartData, setChartData] = useState([]);
+  const [legendData, setLegendData] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [total, setTotal] = useState(0);
+  const [resultAggregatorError, setResultAggregatorError] = useState(false);
+  const [days, setDays] = useState(params.days);
```
mshriver, line 40: Pull chart_type out into state as well

frontend/src/widgets/resultaggregator.js
```diff
@@ -22,176 +22,164 @@ import WidgetHeader from '../components/widget-header';
...
+  const [groupField, setGroupField] = useState(params.group_field);
 
-  getResultData () {
-    this.setState({isLoading: true});
-    HttpClient.get([Settings.serverUrl, 'widget', 'result-aggregator'], this.params)
+  useEffect(() => {
+    setIsLoading(true);
+    HttpClient.get([Settings.serverUrl, 'widget', 'result-aggregator'], params)
```
mshriver, line 45: Compose the object for query string

`{chart_type: chart_type, group_field: group_field, days: days}`

To exclude params from dependencies list - but you'll have to spread any of the other fields into state too.

mshriver (approved):

PR #573, opened 2025-02-18
Convert UserTokens

mshriver:
syntax updates to AddTokenModal

simplification of tokens page, remove pagination from URL params

### DIFF
--- a/frontend/src/pages/profile/tokens.js
+++ b/frontend/src/pages/profile/tokens.js
@@ -1,6 +1,5 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import { ToastContainer, toast } from 'react-toastify';
+import React, { useEffect, useState } from 'react';
+import { ToastContainer } from 'react-toastify';
 
 import {
   Button,
@@ -11,8 +10,8 @@ import {
   FlexItem,
   PageSection,
   PageSectionVariants,
-  Text,
   TextContent,
+  Title,
 } from '@patternfly/react-core';
 import { PlusCircleIcon } from '@patternfly/react-icons';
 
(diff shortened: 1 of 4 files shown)


frontend/src/pages/profile/tokens.js
```diff
@@ -86,134 +47,102 @@ export class UserTokens extends React.Component {
...
-      .then(data => this.setState({
-        rows: data.tokens.map((token) => this.tokenToRow(token)),
-        page: data.pagination.page,
-        pageSize: data.pagination.pageSize,
-        totalItems: data.pagination.totalItems,
-        totalPages: data.pagination.totalPages,
-        isEmpty: data.pagination.totalItems === 0
-      }))
+      .then(data => {
+        if (data?.tokens) {
```
LightOfHeaven1994, line 65: If there are to tokens we will get here anyway because `data?.tokens` is true condition. We should probably have here `data?.tokens.length > 0`
Otherwise we will not see correct empty state when we have 0 tokens

frontend/src/pages/profile/tokens.js
```diff
@@ -86,134 +47,102 @@ export class UserTokens extends React.Component {
...
+            </Button>
+          </FlexItem>
+        </Flex>
+      </PageSection>
+      <PageSection>
+        <Card>
+          <CardBody className="pf-u-p-0">
+            <FilterTable
+              columns={COLUMNS}
+              rows={tokens ? tokens.map((t) => tokenToRow(t)) : [getSpinnerRow(4)]}
```
LightOfHeaven1994, line 112: getSpinnerRow is not reachable 
```
> [] ? 'dont show spinner' : 'show spinner'
< 'dont show spinner'
```
To fix loading state rows={tokens}

frontend/src/pages/profile/tokens.js
```diff
@@ -22,62 +21,24 @@ import { getSpinnerRow } from '../../utilities';
...
 
-export class UserTokens extends React.Component {
+const UserTokens = () => {
 
-  static propTypes = {
-    location: PropTypes.object,
-    navigate: PropTypes.func,
-    eventEmitter: PropTypes.object
-  };
+  const [tokens, setTokens] = useState([]);
```
LightOfHeaven1994, line 30: To fix loading state `const [tokens, setTokens] = useState([getSpinnerRow(4)]);`

frontend/src/pages/profile/tokens.js
```diff
@@ -86,134 +47,102 @@ export class UserTokens extends React.Component {
...
-        page: data.pagination.page,
-        pageSize: data.pagination.pageSize,
-        totalItems: data.pagination.totalItems,
-        totalPages: data.pagination.totalPages,
-        isEmpty: data.pagination.totalItems === 0
-      }))
+      .then(data => {
+        if (data?.tokens) {
+          setIsEmpty(false);
+          setTokens(data.tokens);
```
LightOfHeaven1994, line 67: To fix loading state `setTokens(data.tokens.map((t) => tokenToRow(t)));`

LightOfHeaven1994 (commented):
Generally LGTM, couple suggestions to fix empty state and loading state

frontend/src/pages/profile/tokens.js
mshriver: ack on the logic here, thanks for checking. javascript truthiness is the stupidest of truthiness.

frontend/src/pages/profile/tokens.js
mshriver: changing the logic, I moved this into the JSX so that `tokens` doesn't hold a list of not tokens.

frontend/src/pages/profile/tokens.js
mshriver: Given the above intent about tokens only having tokens - I think you're right to actually fix the empty/loading state it needs to be initialized in state.

frontend/src/pages/profile/tokens.js
```diff
@@ -86,134 +47,102 @@ export class UserTokens extends React.Component {
...
+            </Button>
+          </FlexItem>
+        </Flex>
+      </PageSection>
+      <PageSection>
+        <Card>
+          <CardBody className="pf-u-p-0">
+            <FilterTable
+              columns={COLUMNS}
+              rows={tokens.map((t) => tokenToRow(t))}
```
LightOfHeaven1994, line 112: @mshriver it will not work with initial state defined as [getSpinnerRow(4)] as it will try to apply tokenToRow helper functon on it. Currently if you load the page - you will not see spinner. it might be not visible as it happens fast, but to test it you can put `debugger;` before HTTP request here `HttpClient.get([Settings.serverUrl, 'user', 'token'], {page: page, pageSize: pageSize})`

PR #576, opened 2025-02-18
Convert EmptyObject

mshriver:


LightOfHeaven1994 (approved):
LGTM, there is 1 issue that is not related to this PR at all. NavLink says that it will redirect you to entities list page but it has wrong link. We may want to fix it later

PR #578, opened 2025-02-25
Convert resultSummary Widget

Fynardo:


frontend/src/widgets/resultsummary.js
```diff
@@ -17,132 +17,127 @@ import { Settings } from '../settings';
...
       })
-      .then(data => this.setState({summary: data}))
+      .then(data => {
+        setSummary(data);
+        setIsLoading(false);
+        setDataError(false);
+      })
       .catch(error => {
-        this.setState({dataError: true});
+        setDataError(true);
```
mshriver, line 59: `setIsLoading(false)` here too? 

In either case, it's done "loading"

PR #593, opened 2025-03-05
Convert GenericBar Widget

hugohezel:
**Note**: branch's name should point to the IQE-3351

### DIFF
--- a/frontend/src/widgets/genericbar.js
+++ b/frontend/src/widgets/genericbar.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
 
 import {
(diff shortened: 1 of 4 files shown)


frontend/src/widgets/genericbar.js
```diff
@@ -21,73 +21,66 @@ import { Settings } from '../settings';
...
       .catch(error => {
-        this.setState({genericBarError: true});
+        setGenericBarError(true);
         console.log(error);
       });
   }
 
-  getBarStyle = (key) => {
+  const onGroupFieldSelect = (value) => {
+    params.group_field = value;
```
mshriver, line 76: Note how I've separated these things into state in #592, since they can initially be passed as parameters but will be modified by the user and should be tracked in state instead of mutating the parameter.

frontend/src/widgets/genericbar.js
```diff
@@ -105,172 +98,181 @@ export class GenericBarWidget extends React.Component {
...
-          <Chart
-            domainPadding={ this.props.horizontal ? { x: 20 } : { y: 20} }
-            padding={ this.props.padding || {
-              bottom: 30,
-              left: 150,
-              right: 15,
-              top: 20
+  return (
+    <Card>
+      <WidgetHeader title={title} getDataFunc={getData} onEditClick={onEditClick} onDeleteClick={onDeleteClick}/>
```
mshriver, line 191: You can drop passing `getDataFunc` for this, I plan to drop it from WidgetHeader.  It was necessary when you couldn't reload the page because there was not static pathing.  Now you can reload the page, instead of trying to individually reload some widget's content.

This allows for making `getData` a `useEffect` function itself.

mshriver (approved):
Testing looks great!  Just clean up the yarn lint warning for react please.

PR #594, opened 2025-03-05
Convert Ibutsu Page

hugohezel:


frontend/src/components/ibutsu-page.js
```diff
@@ -21,54 +21,50 @@ import { ToastContainer } from 'react-toastify';
 import { ALERT_TIMEOUT } from '../constants';
 
 
+const IbutsuPage = (props) => {
+  const {
+    eventEmitter,
```
mshriver, line 26: I'm going to follow this up to remove the eventEmitter completely, but for now it's still connected to the sidebar function call.

frontend/src/components/ibutsu-page.js
```diff
@@ -21,54 +21,50 @@ import { ToastContainer } from 'react-toastify';
...
+      <ToastContainer stacked autoclose={ALERT_TIMEOUT} />
+    </React.Fragment>
+  );
 
-  }
+};
 
+IbutsuPage.propTypes = {
+  eventEmitter: PropTypes.object,
+  navigation: PropTypes.node,
```
mshriver, line 63: navigation, location, and params are no longer props used by the component.

mshriver: Going to merge the commit as-is since I'm going to follow it anyway with removal of eventEmitter

hugohezel: ACK, I forgot about these, thanks

PR #612, opened 2025-03-24
fix(IQE-3443): Apply current selection to the Dashboard selector

hugohezel:
## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the filter value was not being updated correctly when a dashboard was selected.


hugohezel:
> * Consider adding a comment explaining why `setFilterDBValue` needs to be called with the title.

Without this, there is an edge case in which the reported behaviour is present, when the page is refreshed.

mshriver (approved):

PR #617, opened 2025-04-01
Fix results table visual bugs

Fynardo:
This PR fixes a couple of visual bugs in Results table.
* Add a spinner to let users know that the table is loading results
* Migrate resultSelection from list to text so the placeholdet text shows correctly

Closes IQE-3442

## Summary by Sourcery

Improve user experience in the Results table by adding a loading state and fixing a placeholder text issue

New Features:
- Add a loading spinner to indicate when results are being fetched

Bug Fixes:
- Fix placeholder text display for result selection

Enhancements:
- Implement a more informative loading state for the results table

### DIFF
--- a/frontend/src/result-list.js
+++ b/frontend/src/result-list.js
@@ -239,6 +239,7 @@ const ResultList = () => {
   useEffect(() => {
     const fetchData = async () => {
       setIsError(false);
+      setRows([getSpinnerRow(5)]);
       let apiParams = { filter: [] };
       let newFilters = { ...filters };
       if (primaryObject) {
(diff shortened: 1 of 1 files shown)


frontend/src/result-list.js
```diff
@@ -72,7 +72,7 @@ const ResultList = () => {
   const [runInputValue, setRunInputValue] = useState('');
   const [runFilterValue, setRunFilterValue] = useState('');
 
-  const [resultSelection, setResultSelection] = useState([]);
+  const [resultSelection, setResultSelection] = useState('');
```
mshriver, line 75: I'm not confident this is the right choice here, since multiple functions in the module treat this as a list.

Consider instead changing the single line that expects a string to check for an empty list:
```
      {resultSelection.length !== 0 ? resultSelection : 'Select a result'}
```

PR #622, opened 2025-04-17
fix(IQE-3502): Be able to trim slashes from both begging and end in URL building

hugohezel:
## Summary by Sourcery

Bug Fixes:
- Update trim function to correctly remove leading and trailing slashes from URLs by modifying the input string in-place

### DIFF
--- a/frontend/src/services/http.js
+++ b/frontend/src/services/http.js
@@ -3,10 +3,10 @@ import { AuthService } from './auth';
 
 const trim = (string) => {
   if (string.startsWith('/')) {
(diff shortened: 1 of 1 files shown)


frontend/src/services/http.js
```diff
@@ -3,10 +3,10 @@ import { AuthService } from './auth';
 
 const trim = (string) => {
   if (string.startsWith('/')) {
-    return string.slice(1);
```
mshriver, line 6: Good catch, end up returning without trimming the trailing slash when this returns.

PR #628, opened 2025-04-22
Use an object for UserFilter options

mshriver:
Provide a display name for the filter options instead of the api parameter key


mshriver:
@sourcery-ai review

PR #629, opened 2025-04-23
feat(IQE-3524): Remove option for selecting project owner in Add Project form

hugohezel:
## Summary by Sourcery

Remove the ability to manually select a project owner when creating a new project

New Features:
- Automatically set the project owner to the user creating the project when adding a new project

Bug Fixes:
- Prevent potential issues with incorrect project owner assignment during project creation

Enhancements:
- Simplified project creation workflow by removing manual owner selection

### DIFF
--- a/frontend/src/pages/admin/project-edit.js
+++ b/frontend/src/pages/admin/project-edit.js
@@ -56,6 +56,9 @@ const ProjectEdit = () => {
   const params = useParams();
   const navigate = useNavigate();
 
+  // is project new or existing
+  const [isNewProject, setIsNewProject] = useState(false);
+
   // project title and name states, required
   const [id, setId] = useState();
   const [title, setTitle] = useState('');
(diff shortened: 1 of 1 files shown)


frontend/src/pages/admin/project-edit.js
```diff
@@ -329,42 +329,44 @@ const ProjectEdit = () => {
                   </HelperText>
                 </FormHelperText>
               </FormGroup>
-              <FormGroup fieldId="owner" label="Owner">
```
mshriver, line 332: I would prefer that we keep the FormGroup, and set `isDisabled` for the toggleOwner memo.

https://v5-archive.patternfly.org/components/menus/menu-toggle#menutoggle

Paired with updating the `HelperTextItem` content, you can explain that owner will be assigned to the current user and can be updated after saving.

mshriver (approved):

PR #630, opened 2025-04-23
fix(IQE-3525): Expand the click area to the entire item area in User Dropdown

hugohezel:


mshriver (approved):
Great find!

frontend/src/components/user-dropdown.js
```diff
@@ -48,16 +49,16 @@ const UserDropdown = () => {
       )}
     >
       <DropdownList>
-        <DropdownItem key="profile">
-          <Link to="/profile/user">Profile</Link>
+        <DropdownItem key="profile" onClick={() => navigate('/profile/user')}>
```
mshriver, line 52: Alternatively, you could use `onSelect` for the `Dropdown` to navigate based off the value

PR #657, opened 2025-05-29
Debounce TestHistory requests

mshriver:
Opening PR early with debug branch content.

#656 Moved the update for the `result` filter state out of the handler, instead only setting it once in the effect and adding the dependency on `onlyFailures` selection. This may be part of the root cause of the behavior and is being deployed to stage now.

## Summary by Sourcery

Debounce repeated fetch calls in TestHistoryTable and ResultAggregatorWidget to prevent backend hangs and update the time‐range filter logic to correctly depend on selectedTimeRange.

Bug Fixes:
- Prevent backend hangs by debouncing getResults and resultAggFetch calls in TestHistoryTable

Enhancements:
- Standardize pagination state to use numeric page and pageSize across multiple components
- Refactor TestHistory time‐range selection to rely on selectedTimeRange and close the selector without rebuilding active filters directly
- Allow days prop in ResultAggregatorWidget to accept both strings and numbers
- Remove unused debounce utility function

### DIFF
--- a/frontend/src/components/classify-failures.js
+++ b/frontend/src/components/classify-failures.js
@@ -110,8 +110,8 @@ const ClassifyFailuresTable = () => {
         );
         const data = await HttpClient.handleResponse(response);
         setFilteredResults(data.results);
-        setPage(data.pagination.page.toString());
-        setPageSize(data.pagination.pageSize.toString());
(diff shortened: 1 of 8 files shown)


mshriver:
@sourcery-ai review

PR #663, opened 2025-06-03
Rework Chart colors

### LINKED ISSUE #463: [Bug] Pie Chart widget shows incorrect legend colors and values
While adding an Aggregation Result Widget(pie chart) I came across incorrect legends and values for the same. 
If we see the pie chart of screenshot 1, the "Stage" legend has been shown twice in different colors and they do not match where I have grouped data by `env`
Additionally there is another scenario where I try to group data by `result` it shows the legends color incorrectly

image

Screenshot from 2023-06-06 17-40-50

mshriver:
Use data styles and legendData, along with a new constant for chart colors

Use chart colors instead of global colors for charts

Some async/await patterns too

FIXES #463

## Summary by Sourcery

Rework chart color management and streamline data fetching across various widgets

Enhancements:
- Introduce CHART_COLOR_MAP constant for unified chart colors and update all chart widgets to use it
- Refactor widget data fetching to use async/await with debounced useEffect hooks
- Generate legendData via useMemo and replace hard-coded colorScale and getColors props in bar, area, heatmap and summary widgets
- Add HEATMAP_TYPES constant and refactor FilterHeatmapWidget to fetch analysis view ID and link using hooks
- Implement loadKey with nanoid in Dashboard and useWidgets to trigger widget reloads on add/edit/delete
- Update accessibility analysis view to use themeColor for ChartLegend ordering

Documentation:
- Clarify default "chart_type" description to include "donut" in backend constants

### DIFF
--- a/frontend/src/dashboard.js
+++ b/frontend/src/dashboard.js
@@ -38,6 +38,8 @@ import DeleteModal from './components/delete-modal';
 import { useWidgets } from './components/hooks/useWidgets';
 import { IbutsuContext } from './components/contexts/ibutsuContext';
 
+import { nanoid } from 'nanoid/non-secure';
+
 const Dashboard = () => {
   const { defaultDashboard, primaryObject } = useContext(IbutsuContext);
   const { dashboard_id, project_id } = useParams();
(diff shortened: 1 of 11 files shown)


frontend/src/dashboard.js
```diff
@@ -64,6 +66,7 @@ const Dashboard = () => {
   const [selectInputValue, setSelectInputValue] = useState('');
   const [selectFilterValue, setSelectFilterValue] = useState('');
   const selectInputRef = useRef();
+  const [loadKey, setLoadKey] = useState(nanoid(6));
```
mshriver, line 69: This is not a very efficient way to do this, at all, but given the scale of the number of widgets on dashboards, I'm going to go with this for now.

I've worked on making a more targeted update to widgets on edit, but I think it will take significant changes that will go well beyond the intent of this PR. I've been iterating with Sonnet and am not that close to a reasonable solution.

PR #817, opened 2026-03-24
docs(user-guide): JUnit import, import API examples, and widgets page

vishnandaman:
This PR introduces a detailed user guide for Dashboards and Widgets and improves the JUnit XML import documentation. While exploring Ibutsu as a SWE intern, I noticed that the process for setting up widgets and the non-Pytest upload flow could be better documented for new users.

Changes
- Added a detailed curl example for the API import that includes project, source, and metadata flags required for proper dashboard mapping.

New Guide: user-guide/widgets.rst
- Added step-by-step instructions for the Add Widget wizard.
-  Created a list-table of built-in widget types (Heatmaps, Aggregators, etc.) aligned with the app's KNOWN_WIDGETS.

How to verify
I have verified the build locally using the following steps:

cd docs
python3 -m venv .venv && source .venv/bin/activate
pip install -r requirements.txt
make html

## Summary by Sourcery

Document dashboard widgets and improve JUnit XML import guidance in the user guide.

Documentation:
- Add a new widgets user guide with step-by-step instructions for using the Add Widget wizard and descriptions of built-in widget types.
- Expand the JUnit XML import documentation with a detailed curl example that covers project, source, and metadata parameters.
- Update related user guide navigation and references to surface the new widgets and import documentation.


docs/source/user-guide/getting-started.rst
```diff
@@ -91,7 +101,16 @@ one may use the following curl command:
 
 .. code-block:: shell
 
-  curl --form importFile=@<file-name>.xml http://ibusu-server/api/import
+   curl --request POST \
```
mshriver, line 104: I think it would be better for us to recommend the API clients for python and typescript here:
https://github.com/ibutsu/ibutsu-client-python
https://github.com/ibutsu/ibutsu-client-javascript

I think it's good to continue to include this raw curl example, but first recommendation should be to use the client in python/typescript. Those libraries include usage examples and I don't think you need to repeat them here.

mshriver (approved):
Comment about api clients can be added in followup commit, it's non-blocking.

PR #833, opened 2026-04-14
Bug fixes

mshriver:
## Summary by Sourcery

Fix artifact and project access control edge cases and improve robustness around user/project lookups.

Bug Fixes:
- Ensure artifact access checks correctly handle artifacts linked via either results or runs and deny access to orphaned artifacts.
- Prevent errors when project_has_user is called with None or a non-existent project ID by explicitly returning False in those cases.
- Correct project permission checking during import to use the resolved project object instead of the raw identifier.
- Fix password recovery activation_code to be stored as a string rather than bytes.

Enhancements:
- Refactor artifact access validation into a shared helper to centralize project-based permission logic.

Tests:
- Add tests covering project_has_user behavior when given None or string IDs that do not resolve to a project.

### DIFF
--- a/backend/ibutsu_server/controllers/artifact_controller.py
+++ b/backend/ibutsu_server/controllers/artifact_controller.py
@@ -82,6 +82,8 @@ def view_artifact(id_, token_info=None, user=None):
     :rtype: file
     """
     artifact, response = _build_artifact_response(id_)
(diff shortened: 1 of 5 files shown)


mshriver:
@sourcery-ai review

PR #1010, opened 2026-08-03
Handle header column accessibility

mshriver:
## Summary by Sourcery

Improve accessibility handling for table header columns, especially when column titles are empty or represented as objects.

Enhancements:
- Introduce shared helpers to derive column text and detect empty header columns across the filtered table component.
- Ensure skeleton table columns mirror real table accessibility by adding screen-reader-only labels for empty headers while preserving other column properties.


mshriver:
@sourcery-ai review

